PR #4, opened 2016-01-14
Fix today date styling

rishigoutam:
Fix isToday check so that .currentday class isn't applied to days in months other than the current month.


Ashoat:
In favor of https://github.com/Ashoat/squadcal/pull/2/commits/0631db75231cb9d15727bc3055d3608943830c13

PR #5, opened 2016-01-14
Add selection for current selected date

rishigoutam:
Since the date is written in the lower right of the textbox, it's hard to determine which box is selected. A user might end up editing the wrong box. Adding some css will make it more obvious which date is being edited.

### DIFF
--- a/index.php
+++ b/index.php
@@ -47,6 +47,9 @@ while ($row = $result->fetch_assoc()) {
                   overflow-y: hidden;
                   padding: 5px;
                 }
+                textarea.selected {
+                  border: 1px dotted;
+                }
                 * {
                   padding: 0;
                   margin: 0;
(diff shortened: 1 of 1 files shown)


Ashoat:
I'll take it but tbh bad color, will fix later

Ashoat:
Also see http://www.w3schools.com/cssref/sel_active.asp

rishigoutam:
Sorry for the color. I'm color blind :(

Ashoat:
See here for fixes: https://github.com/Ashoat/squadcal/commit/10e3aecb04281c95a3eb14a7f447a48bf2a4651c

1) CSS can do it all, no JS necessary
2) You want to offset the increase in the border by a decrease in padding to avoid text getting shifted when it's clicked on
3) Color please
4) The date looks a little close to the border when it's clicked, so I moved it a bit

PR #46, opened 2020-06-01
[native] Friend list / block list screens, add friend / block friend modal

zrebcu411:


lib/actions/user-actions.js
```diff
@@ -290,6 +290,18 @@ async function requestAccess(
...
 }
 
+const sendFriendRequestActionTypes = Object.freeze({
+  started: 'SEND_FRIEND_REQUEST_STARTED',
+  success: 'SEND_FRIEND_REQUEST_SUCCESS',
+  failed: 'SEND_FRIEND_REQUEST_FAILED',
+});
+async function sendFriendRequest(
+  fetchJSON: FetchJSON,
+  userIDs: string[],
```
Ashoat, line 300: Nit: `$ReadOnlyArray<string>`

lib/types/account-types.js
```diff
@@ -148,3 +148,7 @@ export type AccessRequest = {|
   email: string,
   platform: DeviceType,
 |};
+
+export type AddFriendsRequest = {|
+  userIDs: string[],
```
Ashoat, line 153: Nit: `$ReadOnlyArray<string>`

lib/types/redux-types.js
```diff
@@ -792,6 +792,22 @@ export type BaseAction =
...
       payload: LocallyComposedMessageInfo,
+    |}
+  | {|
+      type: 'SEND_FRIEND_REQUEST_STARTED',
+      payload?: void,
+      loadingInfo: LoadingInfo,
+    |}
+  | {|
+      type: 'SEND_FRIEND_REQUEST_SUCCESS',
+      payload: {},
```
Ashoat, line 803: Just a heads-up, `{}` is actually pretty unsound in Flow (as is `Object`). Right now it looks like the payload is `void` (which is `typeof undefined`), but if you're trying to type an object, some options are `{||}` (enforced empty object) or `{| +[key: string]: mixed |}` (generic object with read-only properties)

native/more/add-friends-modal.react.js
```diff
@@ -0,0 +1,309 @@
+// @flow
+
+import React from 'react';
```
Ashoat, line 3: Nit: I prefer to import everything `import * as React from 'react';`, open to being convinced otherwise though

native/more/add-friends-modal.react.js
```diff
@@ -0,0 +1,309 @@
...
+import { Text, View, ActivityIndicator, Alert } from 'react-native';
+import { createSelector } from 'reselect';
+import invariant from 'invariant';
+
+import type { RootNavigationProp } from '../navigation/root-navigator.react';
```
Ashoat, line 8: Nit - I usually separate imports into four ordered categories:
1. Types
2. Imports from outside monorepo
3. Imports from other packages inside monorepo (usually `lib`)
4. Imports from the local package

Open to changing this, though. What do you think? Alternately, we could just set up Prettier/ESLint to auto-alphabetize the imports or something

native/more/friend-list-item.react.js
```diff
@@ -0,0 +1,124 @@
...
+  type KeyboardState,
+  withKeyboardState,
+} from '../keyboard/keyboard-state';
+import {
+  withOverlayContext,
+  type OverlayContextType,
+} from '../navigation/overlay-context';
+import { type Colors, colorsSelector, styleSelector } from '../themes/colors';
+
+// TODO: move pencil icon up to components?
```
Ashoat, line 27: Yes, good idea :)

native/more/friend-list.react.js
```diff
@@ -0,0 +1,64 @@
...
+
+const styles = {
+  container: {
+    flex: 1,
+  },
+};
+const stylesSelector = styleSelector(styles);
+
+export default connect((state: AppState) => ({
+  styles: stylesSelector(state),
```
Ashoat, line 63: The point of `stylesSelector` is actually only about color theming. It basically just replaces any of the strings in `native/themes/colors.js` with the corresponding color depending on whether the user is in dark mode or light mode. In this case you're not using any colors, so I would avoid using Redux and just reference the `styles` object directly (unless you have plans to add colors to `styles`)

server/src/responders/user-responders.js
```diff
@@ -335,6 +336,24 @@ async function requestAccessResponder(
...
+const friendRequestInputValidator = tShape({
+  userIDs: t.list(t.String),
+});
+
+async function friendRequestCreationResponder(
+  viewer: Viewer,
+  input: any,
+): Promise<void> {
+  const request: AddFriendsRequest = input;
+  await validateInput(viewer, friendRequestInputValidator, request);
```
Ashoat, line 348: One thing we haven't talked about is the experience of receiving a friend request. I think it should pop up as a new thread, since the app is based on this assumption that all notifs come from threads.

Probably worth doing in a separate PR later. [Here's a link](https://github.com/Ashoat/squadcal/blob/master/server/src/bots/app-version-update.js#L62-L67) to some existing code that does something similar.

I also just created #50 to track the more general question of what a thread with a non-friend looks like.

server/src/responders/user-responders.js
```diff
@@ -335,6 +336,24 @@ async function requestAccessResponder(
...
+
+  return new Promise(resolve => {
```
Ashoat, line 350: Something else we'll probably need on the server: when we launch the feature, we'll probably want to bootstrap the friend lists based on existing connections in threads. This would be a separate PR as well, but basically when we launch the friend list we can consider anybody who is in a thread with you as your friend. We probably would just set up a script in `server/src/scripts`.

I created #51 to track

native/more/friend-list.react.js
zrebcu411: Have you ever considered using one of `css-in-js` libraries like [styled-components](https://styled-components.com/)? It works well on web and react-native, supports theming out of the box, and allows to use css syntax in react-native. JSX is much clearer since you don't have to pass `style` props every time and to have conditions:

```javascript
// App.js

const theme = {
  dark: {
     boxBackground: '#ffffff',
  },
  light: {
     boxBackground: '#000000',
  }
}

function App() {
  return (
     <ThemeProvider theme={dark ? theme.dark : theme.light}>
       <Screen />
     <ThemeProvider>
  )
}
```

```javascript
// Screen.js

import styled, { css } from 'styled-components/native';

...
function Screen(props) {
  return (
     <Box primary onLayout={() => console.log('I accept View props')}>
        <Title>I'm a Text but I have a title styles</Title>
     </Box>
  )
}

const Box = styled.View`
  background-color: ${props => props.theme.boxBackground}; // 1
  border: 1px solid ${props => props.theme.boxBorder}; // 2
  padding: 10px 5px; // 3

  ${props => props.primary && css`
    background-color: ${props => props.theme.boxPrimaryBackground};
    border-color: ${props => props.theme.boxPrimaryBorder};
  `} // 4
`;

// 1. It's easy to interpolate theme provided by ThemeProvider. It can be dynamic
// 2. It allows using web syntax in native for rules like border or box-shadow
// 3. It allows using px units in native which helps with writing responsive views for different device screen sizes
// 4. It allows to pass custom props and to avoid handling conditions in JSX
```

It supports SCSS syntax on the web (it allows to easily override styles of third-party components etc):
``` javascript
import styled, { css } from 'styled-components';

const Container = styled.div`
  background-color: 'blue';

  .inner {
     background-color:  'red';

     p {
        color: 'green';
     }
  }
`

<Container>
   <div class="inner">
      <p>Hello world</p>
   </div>
</Container>
```

zrebcu411:
Thank you for the very first review :) I pushed the updates you mentioned.

I have one question: do we need to write PropTypes if we use FlowType?

native/more/friend-list.react.js
Ashoat: I've considered it before but never made the jump. I think this is a good idea, created #55 to track. If it's easy to set up feel free to do it sooner than later, we can handle refactoring the codebase piece-by-piece

native/more/add-friends-modal.react.js
Ashoat: Good question. So it's worth noting that since we're using `babel-eslint`, any `import type` statements get stripped before ESLint sees them. So there are some examples of duplicate `import`s where one is just `import type`.

I don't have a strong opinion on how to handle types vs. normal imports, you'll see me do it in both ways across the codebase. I'm open to whatever your preference is, we can standardize on that. But in general I think my approach has been - if it's a small type that is connected to the import, then do it on the same line. But if there are a lot of type imports, or if the type imports aren't necessarily related to the other imports, I sometimes put them on a separate line.

Ashoat:
You can skip the `PropTypes`. I have a lot of them throughout the codebase... they occasionally catch bugs for me but I don't think they are really worth it in a statically-checked codebase. We should maybe just get rid of them at some point in the future.

Ashoat:
By the way, don't worry about the CLA - the agreement I've signed with Software Mansion should cover this. I just need to figure out how to put an exception into @CLAassistant.

native/more/add-friends-modal.react.js
zrebcu411: The way you have been doing it so far is ok for me

native/more/relationship-list-item.react.js
```diff
@@ -0,0 +1,111 @@
...
+import { type Colors, colorsSelector, styleSelector } from '../themes/colors';
+
+type Props = {|
+  userInfo: UserInfo,
+  // Redux state
+  removeUserLoadingStatus: LoadingStatus,
+  colors: Colors,
+  styles: typeof styles,
+  // withKeyboardState
+  keyboardState: ?KeyboardState,
```
Ashoat, line 30: Many of these props aren't being used

native/more/relationship-list.react.js
```diff
@@ -0,0 +1,123 @@
...
+  { id: '2', username: 'Emma' },
+];
+
+class RelationshipList extends React.PureComponent<Props, State> {
+  render() {
+    return (
+      <View style={this.props.styles.container}>
+        <FlatList
+          contentContainerStyle={this.props.styles.contentContainer}
+          ListHeaderComponent={
```
Ashoat, line 39: I'm not sure, but I think right now this is forcing the `FlatList` to rerender everything every time `RelationshipList` is rendered. Besides that it feels a bit icky to have a huge prop like this defined inline.

Some thoughts:

1. Probably the easiest option for now would be to define a function component for this
2. But that wouldn't work well when `data` changes since you need access to that state. You could instead create a memoized render function that returns `ListHeaderComponent` that only changes when `DATA.length !== 0` changes
3. But on a higher level, when you find yourself defining a `ListHeaderComponent` that is actually dynamic, that should be a strong signal that this component should be defined as a normal `data` row
4. Even higher level: I'm not sure this is the best place to put the button for this. What if we placed it in `headerRight`, similar to where `ComposeThreadButton` is on `MessageList`?

native/more/relationship-list.react.js
```diff
@@ -0,0 +1,123 @@
...
+import { type UserInfo } from 'lib/types/user-types';
+
+import { AddFriendsModalRouteName } from '../navigation/route-names';
+import { styleSelector } from '../themes/colors';
+
+import RelationshipListItem from './relationship-list-item.react';
+import RelationshipListAddButton from './relationship-list-add-button.react';
+
+type Props = {|
+  navigation: MoreNavigationProp<'FriendList'>,
```
Ashoat, line 20: `RelationshipList` is used for two different components, but here you've defined it specifically for `FriendList`. In this case you aren't using the navigation params, so it should be fine to just leave the route unspecified instead:

```js
  navigation: MoreNavigationProp<>,
  route: NavigationRoute<>,
```

Actually, I've realized that `NavigationRoute` doesn't have a default type param. For this to work, you'll have to update it to have a default type parameter `$Keys<ScreenParamList>`, similar to how `MoreNavigationProp` works.

lib/actions/user-actions.js
```diff
@@ -290,6 +290,18 @@ async function requestAccess(
   await fetchJSON('request_access', accessRequest);
 }
 
+const sendFriendRequestActionTypes = Object.freeze({
+  started: 'SEND_FRIEND_REQUEST_STARTED',
+  success: 'SEND_FRIEND_REQUEST_SUCCESS',
+  failed: 'SEND_FRIEND_REQUEST_FAILED',
+});
+async function sendFriendRequest(
```
Ashoat, line 298: Do you think it would be better to have the server endpoint be more generic? If the backend treats `FRIEND` as a generic case of a relationship, and returns `UserInfo` objects with a field that says `FRIEND`, maybe we could have a single endpoint for handling both friend requests and blocks

native/more/add-friends-modal.react.js
```diff
@@ -0,0 +1,308 @@
...
+      {
+        userInfoInputArray: [],
+        usernameInputText: '',
+      },
+      this.onErrorAcknowledged,
+    );
+  };
+
+  close = () => {
+    this.props.navigation.goBack();
```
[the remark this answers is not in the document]
zrebcu411: I'm not able to see PR you mentioned. I'm getting 404 🤔

native/more/relationship-list.react.js
zrebcu411: I left it unspecified, but I think it's not a good idea for `route` prop. It would work for `navigation` since `routeName` must be always provided when `navigation.navigate()` is called so Flow can infer the `params` type for that route and keep it strict. However, `route` prop is used for getting the navigation params for the current screen so when we would have `NavigationRoute<>` with the unspecified route name it wouldn't be possible to infer the `params` type. `props.route.params` will be `any` in that case. I don't like having unsafe types in the code.

I'm gonna use RelationshipList as a component that would be used in two separate screen components: FriendListScreen and BlockListScreen. What do you think?

native/more/relationship-list.react.js
zrebcu411: I agree that it would be better to keep it as a memoized function component. And you're 100% right, it would be much better to put it in the header. I will do that

native/more/add-friends-modal.react.js
```diff
@@ -0,0 +1,308 @@
...
+  get userSearchResults() {
+    return this.userSearchResultsSelector({ ...this.props, ...this.state });
+  }
+
+  render() {
+    const loading = this.props.addFriendsLoadingStatus === 'loading';
+
+    let addButton = null;
+    const inputLength = this.state.userInfoInputArray.length;
+    if (inputLength > 0) {
```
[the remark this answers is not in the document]
zrebcu411: I've pushed an update which disables editing

lib/actions/user-actions.js
zrebcu411: I agree. I'm going to create a generic endpoint. That was a temporary solution, just wanted to play a bit with client-server communication. I think this endpoint should not be included in this PR since it's all about UI. I can remove it for now

native/more/relationship-list-add-button.react.js
```diff
@@ -0,0 +1,113 @@
...
+import { connect } from 'lib/utils/redux-utils';
+
+import Button from '../components/button.react';
+import { styleSelector, colorsSelector, type Colors } from '../themes/colors';
+
+// Condider moving to components and creating SectionList/SettingsList UI kit
+// for consistency, reusability and to avoid styling everytime list is used
+// but with keeping it highly customizable for special use cases
+//
+// Declarative API proposal:
```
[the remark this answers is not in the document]
zrebcu411: I've removed this comment and created an issue: https://github.com/Ashoat/squadcal/issues/64

native/more/more.react.js
```diff
@@ -39,11 +44,28 @@ const deleteAccountOptions = { headerTitle: 'Delete account' };
 const buildInfoOptions = { headerTitle: 'Build info' };
 const devToolsOptions = { headerTitle: 'Developer tools' };
 const appearanceOptions = { headerTitle: 'Appearance' };
+const friendListOptions = ({ navigation }) => ({
+  headerTitle: 'Friend list',
+  // eslint-disable-next-line react/display-name
```
zrebcu411, line 49: I don't know if we want to have this ignore line here

Ashoat: Yeah it's correct. It's a quirky situation where `headerRight` is actually never rendered as a `<ReactComponent>`, it's just called like a function. For that reason, setting the `displayName` property on it has no effect. `react/display-name` assumes it's a React function component because it's defined in exactly the same way. In fact, it could be used as a component. But it just isn't

native/more/add-friends-modal.react.js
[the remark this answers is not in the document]
Ashoat: Oops, meant to link https://github.com/Ashoat/squadcal/commit/ab3494fc282df6fc685607110e7fc7a99a4fb0a0

native/navigation/route-names.js
```diff
@@ -97,6 +101,8 @@ export type MoreParamList = {|
   BuildInfo: void,
   DevTools: void,
   AppearancePreferences: void,
+  FriendList: { a: boolean },
```
Ashoat, line 104: What's this for?

lib/actions/user-actions.js
Ashoat: Is the plan to remove the endpoint then?

lib/actions/user-actions.js
```diff
@@ -290,6 +290,18 @@ async function requestAccess(
...
+  fetchJSON: FetchJSON,
+  userIDs: $ReadOnlyArray<string[]>,
```
Ashoat, line 300: This is an array of arrays

native/more/add-friends-modal.react.js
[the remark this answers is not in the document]
Ashoat: Nice, awesome!

native/more/relationship-list.react.js
Ashoat: 1. You're confusing a couple different `RouteName`s. The `DestinationRouteName` in `navigate` is parameterized at invocation, and can be any route name in `GlobalParamList`. The `RouteName` that `NavigationProp` is parameterized on is the `RouteName` of the current route, and it is used only to type `setParams`. In our case we aren't using `setParams`, so it doesn't matter.
2. Are you sure `this.props.route.params` is `any`? I think it should be the union of all the params for every screen in `MoreParamList`. I think it should be sound, but Flow has a lot of quirks - let me know if you're seeing it be typed as `any`. (You can examine the type of a specific property by extracting it `const { params } = this.props.route;` and then using `flow type-at-pos` on it)
3. Using `RelationshipList` for two screens sound like a fine plan. If you end up needing params, as long as they're the same between the two screens it should still be possible to type correctly like `MoreNavigationProp<'FriendList' | 'BlockList'>` I think

native/navigation/route-names.js
zrebcu411: Sorry, it was for testing purposes. I forgot to remove :(

native/more/relationship-list.react.js
zrebcu411: 1. That it what I meant. `navigation.navigate` requires a `routeName` argument so Flow knows what is the type of the `params` of that route based on `GlobalParamList` and `routeName` passed to `navigate` function.
2. You're right. It's a union type. I had to restart flow server to see that. Previously I had no flow errors and `any` for `params`.
3. `MoreNavigationProp<'FriendList' | 'BlockList'>` works fine 👌

zrebcu411:
I removed unfinished server code and added other fixes. Now this PR related only to native part

lib/types/endpoints.js
```diff
@@ -74,6 +74,7 @@ const socketPreferredEndpoints = Object.freeze({
   REQUEST_ACCESS: 'request_access',
   CREATE_MULTIMEDIA_MESSAGE: 'create_multimedia_message',
   DELETE_UPLOAD: 'delete_upload',
+  SEND_FRIEND_REQUEST: 'send_friend_request',
```
Ashoat, line 77: Let's get rid of these too, since we're not planning on actually creating this endpoint

lib/types/redux-types.js
```diff
@@ -792,6 +792,22 @@ export type BaseAction =
   | {|
       type: 'CREATE_LOCAL_MESSAGE',
       payload: LocallyComposedMessageInfo,
+    |}
+  | {|
+      type: 'SEND_FRIEND_REQUEST_STARTED',
```
Ashoat, line 797: Get rid of this

native/more/relationship-list-add-button.react.js
```diff
@@ -0,0 +1,44 @@
...
+import { connect } from 'lib/utils/redux-utils';
+
+import Button from '../components/button.react';
+import { styleSelector, colorsSelector, type Colors } from '../themes/colors';
+
+type ListActionProps = {|
+  onPress: () => void,
+  // Redux state
+  colors: Colors,
+  styles: typeof styles,
```
Ashoat, line 17: There is no reason to use `stylesSelector`, as none of the styles use any of the custom color functionality

native/more/relationship-list.react.js
```diff
@@ -0,0 +1,100 @@
...
+  { id: '2', username: 'Emma' },
+];
+
+class RelationshipList extends React.PureComponent<Props> {
+  render() {
+    return (
+      <View style={this.props.styles.container}>
+        <FlatList
+          contentContainerStyle={this.props.styles.contentContainer}
+          ListEmptyComponent={
```
Ashoat, line 36: I thought we discussed the use of the inline definitions like this - they force the rerender of `FlatList` every time `RelationshipList` is rendered

native/more/relationship-list.react.js
```diff
@@ -0,0 +1,119 @@
...
+      <Text
+        style={this.props.styles.emptyText}
+      >{`You haven't added any users yet`}</Text>
+    );
+
+    return (
+      <View style={this.props.styles.container}>
+        <FlatList
+          contentContainerStyle={this.props.styles.contentContainer}
+          ListEmptyComponent={empty}
```
Ashoat, line 48: Let's just handle this in `mapListData` as well

native/more/relationship-list.react.js
```diff
@@ -0,0 +1,119 @@
...
+          data={listData}
+          renderItem={this.renderItem}
+        />
+      </View>
+    );
+  }
+
+  mapListData = (): ListItem[] => {
+    const users = DATA;
```
Ashoat, line 57: This should be memoized based on the input data, so that it doesn't get recomputed every render. Since the memoization will be done using `reselect` long-term and there's no way to mock fake input to `reselect`, I guess we can do this later...

An alternative would be to make your mocked `DATA` array just be what you pass to `FlatList`. This would mean shelving the `mapListData` code and bringing it back later once you have the actual data in place.

zrebcu411: Done. But I'm a little bit confused. I was trying to do it the same way it is done on [thread-settings](https://github.com/Ashoat/squadcal/blob/master/native/chat/settings/thread-settings.react.js#L318). And looks like `listData` in there is recalculated every render since it happens directly on render method. Even if the selector used in `getThreadInfo` is memoized 🤔

native/more/relationship-list.react.js
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  renderItem = ({ item }: { item: ListItem }) => {
+    if (item.type === 'empty') {
+      return (
+        <Text
+          style={this.props.styles.emptyText}
+        >{`You haven't added any users yet`}</Text>
+      );
+    }
```
Ashoat, line 74: Nit: I prefer `else if` for these type of situations

native/more/more-screen.react.js
```diff
@@ -213,6 +215,26 @@ class MoreScreen extends React.PureComponent<Props> {
               />
             </View>
           </View>
+          <View style={this.props.styles.slightlyPaddedSection}>
```
Ashoat, line 218: I think there may have been a misunderstanding here. Did you uncomment this because you read my longer message and saw I mentioned that I don't like committing commented out code?

It's true I don't like committing "work-in-progress" code to master, and that includes commented out code. But since these buttons leads to work-in-progress functionality, they're definitely something that needs to be hidden in master. The reason is that when I make a new release, this stuff can't be in it. **My preference is simply to delete this sort of code and send it in a future commit, but in this specific case I'm okay with it being commented out.**

Separately, I'd appreciate any advice you have on how I can communicate better. In my last comment, I tried to **emphasize that we didn't need to do any more changes** in anything other than `RelationshipList`, because I was scared we'd up needing more **review cycles**:

> Don't worry about making any changes based on what I say below... I just wanted to make sure I communicate well.

> Let's definitely not worry about changing it now.

Maybe I should've bolded these parts. Sometimes I worry that I'm writing too much, and the important stuff gets lost.

PR #57, opened 2020-06-04
Background threads

KatPo:


lib/shared/thread-utils.js
```diff
@@ -52,10 +52,23 @@ function viewerIsMember(threadInfo: ?(ThreadInfo | RawThreadInfo)) {
...
+  return threadInfo && threadInfo.currentUser.subscription.home;
+}
+
 function threadInChatList(threadInfo: ?(ThreadInfo | RawThreadInfo)) {
   return (
     viewerIsMember(threadInfo) &&
-    threadHasPermission(threadInfo, threadPermissions.VISIBLE)
+    threadHasPermission(threadInfo, threadPermissions.VISIBLE) &&
+    threadIsInHome(threadInfo)
+  );
```
KatPo, line 64: Because of this (for now at least) in web version only visible chats are those not in background

native/chat/chat-thread-list.react.js
```diff
@@ -255,8 +259,17 @@ const styles = {
 };
 const stylesSelector = styleSelector(styles);
 
-export default connect((state: AppState) => ({
+export const HomeChatThreadList = connect((state: AppState) => ({
   chatListData: chatListData(state),
+  searchChatListData: chatBackgroundListData(state).concat(chatListData(state)),
```
Ashoat, line 264: Doing this here means that `searchChatListData` will change every time the `connect` function is processed. The point of using selectors is memoization - meaning that the result will only change when the inputs change.

With this approach, any time anything in the Redux store changes (even unrelated), this component will have to be rerendered, and the result of `listDataSelector` will have to be recalculated. Making all the rest of the selectors useless.

The "obvious" solution here would be to create a new selector that is `chatBackgroundListData` and `chatListData` together. But I'm realizing perhaps my initial suggestion to simply replace the `chatListData` with a filtered one was bad.

Since we need the full `chatListData` for search, I think instead we should actually get rid of `chatBackgroundListData` and leave the `chatListData` as containing everything. Instead, the individual `ChatThreadList` can just have custom filter functions that filter the `chatListData` within the component to only show the relevant threads that match.

You could define `ChatThreadList` as a component that takes a function prop that can filter what it displays by default, and then memoizes the result of applying the function to the `chatListData` (as is currently done in `listDataSelector`).

`BackgroundChatThreadList` and `HomeChatThreadList` will then be defined as components that render a `ChatThreadList` with a custom filter function. You'll need to take care to make sure that filter function isn't recreated on every render. I'd suggest implementing `BackgroundChatThreadList` and `HomeChatThreadList` as hooks and using `useCallback` to define the filter function, but you could also define them as class components, with the filter functions defined as class properties.

Sorry for the wall of text - let me know if I can help explain over video chat!

native/chat/chat-thread-list.react.js
```diff
@@ -255,8 +259,17 @@ const styles = {
...
+  viewerID: state.currentUserInfo && state.currentUserInfo.id,
+  threadSearchIndex: threadSearchIndex(state),
+  styles: stylesSelector(state),
+}))(ChatThreadList);
+
+export const BackgroundChatThreadList = connect((state: AppState) => ({
```
Ashoat, line 270: Small nit - with the exception of files with tons of exports, such as type definition files, I prefer to do all the exports at once at the end of the file, eg. `export { HomeChatThreadList, BackgroundChatThreadList }`

native/chat/chat-thread-list.react.js
```diff
@@ -47,6 +48,7 @@ type Props = {|
   navigation: ChatNavigationProp<'ChatThreadList'>,
```
Ashoat, line 48: This prop is no longer typed correctly. `ChatThreadList` is now a screen of your new navigator. You'll need to define a new `ChatTopTabsNavigationProp`, similar to how `ChatNavigationProp` is defined

native/navigation/route-names.js
```diff
@@ -30,6 +30,8 @@ export const MoreRouteName = 'More';
 export const MoreScreenRouteName = 'MoreScreen';
 export const ChatRouteName = 'Chat';
 export const ChatThreadListRouteName = 'ChatThreadList';
+export const HomeChatThreadListRouteName = 'Home';
+export const BackgroundChatThreadListRouteName = 'Background';
```
Ashoat, line 34: You'll need to create a new `ParamList` for your new navigator in this file, and make sure it's included in the global `ScreenParamList`

native/chat/chat-thread-list.react.js
KatPo: Please take a look at new commit, I'm not exactly sure if I did this correctly

lib/shared/thread-utils.js
KatPo: Sounds good, for now on web it is like before

lib/shared/thread-utils.js
```diff
@@ -52,13 +52,33 @@ function viewerIsMember(threadInfo: ?(ThreadInfo | RawThreadInfo)) {
...
 function threadInChatList(threadInfo: ?(ThreadInfo | RawThreadInfo)) {
   return (
     viewerIsMember(threadInfo) &&
     threadHasPermission(threadInfo, threadPermissions.VISIBLE)
   );
 }
 
+function threadInBackgroundChatList(threadInfo: ?(ThreadInfo | RawThreadInfo)) {
+  return (
+    viewerIsMember(threadInfo) &&
```
Ashoat, line 68: Let's just call into `threadInChatList(threadInfo)` for these first two conditions

lib/shared/thread-utils.js
```diff
@@ -52,13 +52,33 @@ function viewerIsMember(threadInfo: ?(ThreadInfo | RawThreadInfo)) {
...
+    threadHasPermission(threadInfo, threadPermissions.VISIBLE) &&
+    !threadIsInHome(threadInfo)
+  );
+}
+
+function threadInHomeChatList(threadInfo: ?(ThreadInfo | RawThreadInfo)) {
+  return (
+    viewerIsMember(threadInfo) &&
```
Ashoat, line 76: Same here, let's just call into `threadInChatList(threadInfo)` for these first two conditions

native/chat/chat-thread-background.react.js
```diff
@@ -0,0 +1,27 @@
+// @flow
+
+import * as React from 'react';
+import ChatThreadList from './chat-thread-list.react';
+import type { ChatTopTabsNavigationProp } from './chat.react';
+import { threadInBackgroundChatList } from 'lib/shared/thread-utils';
+import { type ThreadInfo, type RawThreadInfo } from 'lib/types/thread-types';
+
+type BackgroundChatThreadListProps = {
+  navigation: ChatTopTabsNavigationProp<'BackgroundChatThreadList'>,
```
Ashoat, line 10: We should avoid introducing new types using the ambiguous `{ }` syntax. You should always prefer either `{| |}` exact or `{ ... }` inexact syntax. More context [here](https://medium.com/flow-type/on-the-roadmap-exact-objects-by-default-16b72933c5cf), but basically `{ }` currently means inexact but will soon mean exact.

Whenever possible exact syntax is preferred, but in this case because we aren't specifying the `route` prop we have to use inexact:

```js
type BackgroundChatThreadListProps = {
  navigation: ChatTopTabsNavigationProp<'BackgroundChatThreadList'>,
  ...
};
```

native/chat/chat-thread-background.react.js
```diff
@@ -0,0 +1,27 @@
...
+};
+export default function BackgroundChatThreadList(
+  props: BackgroundChatThreadListProps,
+) {
+  const filterFunction = React.useCallback(
+    (item: ?(ThreadInfo | RawThreadInfo)) => {
```
Ashoat, line 16: It looks like you've defined a useless function here. Just pass `threadInBackgroundChatList` in directly, it's the same thing

native/chat/chat-thread-background.react.js
```diff
@@ -0,0 +1,27 @@
...
+      return threadInBackgroundChatList(item);
+    },
+    [],
+  );
+  return (
```
Ashoat, line 21: (We still need to do the `setOptions` call to set the title so it includes the unread count, but actually this would be better in a separate PR)

native/chat/chat-thread-home.react.js
```diff
@@ -0,0 +1,25 @@
...
+import type { ChatTopTabsNavigationProp } from './chat.react';
+import { threadInHomeChatList } from 'lib/shared/thread-utils';
+import { type ThreadInfo, type RawThreadInfo } from 'lib/types/thread-types';
+
+type HomeChatThreadListProps = {
+  navigation: ChatTopTabsNavigationProp<'HomeChatThreadList'>,
+};
+export default function HomeChatThreadList(props: HomeChatThreadListProps) {
+  const filterFunction = React.useCallback(
+    (item: ?(ThreadInfo | RawThreadInfo)) => {
```
Ashoat, line 14: Useless function (see above)

native/chat/chat-thread-home.react.js
```diff
@@ -0,0 +1,25 @@
+// @flow
+
+import * as React from 'react';
+import ChatThreadList from './chat-thread-list.react';
+import type { ChatTopTabsNavigationProp } from './chat.react';
+import { threadInHomeChatList } from 'lib/shared/thread-utils';
+import { type ThreadInfo, type RawThreadInfo } from 'lib/types/thread-types';
+
+type HomeChatThreadListProps = {
+  navigation: ChatTopTabsNavigationProp<'HomeChatThreadList'>,
```
Ashoat, line 10: Use inexact syntax `{...}` (see above)

native/chat/chat-thread-list.react.js
```diff
@@ -44,9 +43,13 @@ const floatingActions = [
 type Item = ChatThreadItem | {| type: 'search', searchText: string |};
 
 type Props = {|
-  navigation: ChatNavigationProp<'ChatThreadList'>,
+  navigation: ChatTopTabsNavigationProp<
+    'HomeChatThreadList' | 'BackgroundChatThreadList',
+  >,
+  filterChats: (threadItem: ThreadInfo) => boolean,
```
Ashoat, line 49: Let's call this `filterThreads`

native/chat/chat-thread-list.react.js
```diff
@@ -44,9 +43,13 @@ const floatingActions = [
 type Item = ChatThreadItem | {| type: 'search', searchText: string |};
 
 type Props = {|
-  navigation: ChatNavigationProp<'ChatThreadList'>,
+  navigation: ChatTopTabsNavigationProp<
```
Ashoat, line 46: The code handling `onTabPress` below needs to be changed now. `tabNavigation` used to be `navigation`'s parent, but now it is its grandparent

native/chat/chat.react.js
```diff
@@ -148,17 +153,37 @@ export type ChatNavigationProp<
...
   ChatNavigationProp<>,
 >();
+
+const ChatThreadsTopTab = createMaterialTopTabNavigator();
+const ChatThreadsComponent = () => {
+  return (
+    <ChatThreadsTopTab.Navigator>
+      <ChatThreadsTopTab.Screen
+        name={HomeChatThreadListRouteName}
+        component={HomeChatThreadList}
```
Ashoat, line 172: Reminder - we want to explicitly specify the title/labels here. Right now it's defaulting to using the route name, but that's meant to be an internal detail instead of a UI label

native/chat/chat.react.js
```diff
@@ -148,17 +153,37 @@ export type ChatNavigationProp<
   RouteName: $Keys<ChatParamList> = $Keys<ChatParamList>,
 > = ChatRouterNavigationProp<ScreenParamList, RouteName>;
 
+export type ChatTopTabsNavigationProp<
+  RouteName: $Keys<ChatTopTabsParamList> = $Keys<ChatTopTabsParamList>,
+> = ChatRouterNavigationProp<ScreenParamList, RouteName>;
```
Ashoat, line 158: `ChatRouterNavigationProp` is the wrong type here, you should be importing `MaterialTopTabNavigationProp`. [This](https://medium.com/@ashoat_t/flow-libdefs-for-react-navigation-5-7dcf1f9da8b7) might be worth a read

native/chat/chat.react.js
```diff
@@ -148,17 +153,37 @@ export type ChatNavigationProp<
...
+
 const Chat = createChatNavigator<
```
Ashoat, line 160: The way this code is organized is messy. We should always try to keep relevant code together, but right now your new stuff is defined in the middle of all of the `ChatComponent` stuff. Let's move all of your new stuff to the very top of the file.

native/navigation/route-names.js
```diff
@@ -30,6 +30,9 @@ export const MoreRouteName = 'More';
 export const MoreScreenRouteName = 'MoreScreen';
 export const ChatRouteName = 'Chat';
 export const ChatThreadListRouteName = 'ChatThreadList';
+export const ChatTopTabsThreadListRouteName = 'ChatTopTabsThreadList';
```
Ashoat, line 33: You don't actually use this anywhere

native/navigation/route-names.js
```diff
@@ -89,6 +92,12 @@ export type ChatParamList = {|
   DeleteThread: DeleteThreadParams,
 |};
 
+export type ChatTopTabsParamList = {|
+  ChatTopTabsThreadList: void,
```
Ashoat, line 96: This isn't used anywhere

native/chat/settings/thread-settings.react.js
```diff
@@ -382,6 +388,11 @@ class ThreadSettings extends React.PureComponent<Props, State> {
       key: 'pushNotifs',
       threadInfo,
     });
+    listData.push({
```
Ashoat, line 391: This class has changed since you put the PR up. (The `listData` is now defined in a selector instead of in the `render` function.)

Can you rebase your changes against master?

Ashoat:
One more thing - we'll also need to update `native/navigation/default-state.js`

native/chat/chat-thread-background.react.js
Ashoat: Actually, we don't even need `useCallback` here since the function definition never changes. Sorry for suggesting that, it was my bad. We can just pass `threadInBackgroundChatList` into `ChatThreadList` directly

native/chat/chat-thread-background.react.js
KatPo: Yes, I'm working on that, just didn't want to put everything in one big PR

native/chat/chat.react.js
KatPo: I'll do that in next PR, didn't want to put everything here

native/chat/chat-thread-list.react.js
```diff
@@ -76,18 +80,22 @@ class ChatThreadList extends React.PureComponent<Props, State> {
   flatList: ?FlatList<Item>;
 
   componentDidMount() {
-    const tabNavigation: ?TabNavigationProp<
+    let tabNavigation: ?TabNavigationProp<
```
Ashoat, line 83: The type here is wrong. The typechecker isn't complaining because it doesn't know how to check `dangerouslyGetParent`, but we still shouldn't be typing things incorrectly. The grandparent is a `TabNavigationProp` but the parent is not. Can you use two different variables here, and type each of them correctly?

native/chat/chat-thread-background.react.js
```diff
@@ -0,0 +1,21 @@
+// @flow
+
+import * as React from 'react';
+import ChatThreadList from './chat-thread-list.react';
+import type { ChatTopTabsNavigationProp } from './chat.react';
```
Ashoat, line 5: Nit: let's follow the conventions I use in other files for separating imports. The imports should be in four groups:

1. Types first (from any place)
2. Imports from external packages next (like `react`)
3. Imports from other packages in our monorepo next (eg. imports from `lib`)
4. Finally, imports from files in the current package (eg. imports starting with `..` or `.`)

native/chat/chat-thread-home.react.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+
+import * as React from 'react';
```
Ashoat, line 3: Sort imports

native/chat/chat-thread-list.react.js
```diff
@@ -67,6 +70,7 @@ class ChatThreadList extends React.PureComponent<Props, State> {
     viewerID: PropTypes.string,
     threadSearchIndex: PropTypes.instanceOf(SearchIndex).isRequired,
     styles: PropTypes.objectOf(PropTypes.object).isRequired,
+    filterThreads: PropTypes.func.isRequired,
```
Ashoat, line 73: Can you sort this in the same order as the `Props` type?

native/chat/chat-thread-list.react.js
```diff
@@ -44,9 +43,13 @@ const floatingActions = [
...
 
 type Props = {|
-  navigation: ChatNavigationProp<'ChatThreadList'>,
+  navigation: ChatTopTabsNavigationProp<
+    'HomeChatThreadList' | 'BackgroundChatThreadList',
+  >,
+  filterThreads: (threadItem: ThreadInfo) => boolean,
   // Redux state
   chatListData: $ReadOnlyArray<ChatThreadItem>,
+  searchChatListData: $ReadOnlyArray<ChatThreadItem>,
```
Ashoat, line 52: This prop is never used

native/chat/chat.react.js
```diff
@@ -38,6 +42,30 @@ import ComposeThreadButton from './compose-thread-button.react';
 import MessageListHeaderTitle from './message-list-header-title.react';
 import ThreadSettingsButton from './thread-settings-button.react';
 import { InputStateContext } from '../input/input-state';
+import {
+  createMaterialTopTabNavigator,
+  type MaterialTopTabNavigationProp,
+} from '@react-navigation/material-top-tabs';
```
Ashoat, line 48: Sort imports

native/chat/chat.react.js
```diff
@@ -38,6 +42,30 @@ import ComposeThreadButton from './compose-thread-button.react';
...
+  RouteName: $Keys<ChatTopTabsParamList> = $Keys<ChatTopTabsParamList>,
+> = MaterialTopTabNavigationProp<ScreenParamList, RouteName>;
+
+const ChatThreadsTopTab = createMaterialTopTabNavigator();
+const ChatThreadsComponent = () => {
+  return (
+    <ChatThreadsTopTab.Navigator>
+      <ChatThreadsTopTab.Screen
+        name={HomeChatThreadListRouteName}
+        component={HomeChatThreadList}
```
Ashoat, line 60: As discussed, please specify label through options here. Let's do it in the current PR

native/navigation/default-state.js
```diff
@@ -38,9 +39,12 @@ const defaultNavigationState: StaleNavigationState = {
                 {
                   name: ChatRouteName,
                   state: {
-                    type: 'stack',
+                    type: 'tab',
                     index: 0,
-                    routes: [{ name: ChatThreadListRouteName }],
+                    routes: [
+                      { name: HomeChatThreadListRouteName },
```
Ashoat, line 45: Why did you replace the stack here? `ChatThreadListRouteName` isn't gone, it just has a child now

native/chat/chat-thread-list.react.js
KatPo: forgot to remove it

KatPo:
Hope that everything is types correctly this time!

Ashoat:
Minor note - in the future make sure to include any updates to `yarn.lock` in your PRs!

Issue #37, opened 2020-05-29
Custom message formatting


Ashoat:
We'd like to support several kinds of message formatting:

1. Replies, similar to how Telegram or Facebook Messenger does it today. A reply to a thread uses the standard Markdown syntax for quoting ("> quote")
2. General Markdown syntax, enabling users to format text in messages, with things like headings and bold/italics
3. Code formatting and JSON pretty-printing, so the dev team can communicate better :)

Ashoat:
### `simple-markdown` rules

We discussed today about what subset of `simple-markdown` rules we want to support. Here's what I have in mind:

- `autolink`
- `blockQuote`
- `codeBlock`
- `del` (looks like ~~strikethrough~~, this one is not so important but looks like it should be easy to support)
- `em` (italics)
- `heading` (multiple kinds via `node.level`)
- `inlineCode`
- `link`
- `mailto` (this one should be the same as `link`)
- `strong` (bold)
- `u` (underline)
- `url` (I'm not sure what the difference is between `url` and `link`...)

### Preserving simple messages

We should also make sure that simple messages without markdown syntax should display exactly like before. Make sure to test messages with line breaks, messages with links, etc.

I also noticed `br`, `newline`, `paragraph`, and `text` rules in `simple-markdown`. I'm not sure how `simple-markdown` parses "simple messages", but we may need to implement "rules" for some of these as well in order to make sure simple messages display the same as before.

### How to structure PRs

I'm realizing this will be a bigger task than I thought initially, so we'll probably need multiple PRs. Let's try to keep the PRs as small as possible. To help structure your work here's a possible order of PRs you might take:

1. PR on native to replace `react-native-hyperlink` in `InnerTextMessage` with our new markdown parser. We will only need the `autolink` rule for this one, plus any rules necessary to "preserve simple messages" as discussed above
2. PR on web to do the same - replace `react-linkify` in `TextMessage` with our new markdown parser that can only handle `autolink`
3. PR on native to replace `react-native-hyperlink` use in other places: `RobotextMessage` and `Entry`. `RobotextMessage` will only need ever need the `autolink` rule (even after you add more rules for `InnerTextMessage`), but `Entry` can have the same rules as `InnerTextMessage`
4. PR on web to replace `react-linkify` in `RobotextMessage` with our new markdown parser (similar to `native`, this will only ever require `autolink` rule)
5. PRs on native to add additional rules. You can group these into however many PRs make sense
6. PRs on web to add additional rules

### List

[...]

PR #67, opened 2020-06-09
Add markdown component with simple rules for native

KatPo:
Take a look and see if this is what we want to achieve in #37 , I also included some screenshots


KatPo:
Web version, in which I think only quotes are broken, 
Zrzut ekranu 2020-06-9 o 09 21 03
Zrzut ekranu 2020-06-9 o 09 21 24

KatPo:
Native version, links are broken (we discussed that), and something is broken with the lists. Also, text in code blocks should have different colors with different themes background, now it's the same as all texts, so with dark color we have light text on light background and it's hard to read
Zrzut ekranu 2020-06-9 o 09 32 28
Zrzut ekranu 2020-06-9 o 09 32 46
Zrzut ekranu 2020-06-9 o 09 33 23

KatPo:
Native version without hyperlink, links work then
Zrzut ekranu 2020-06-9 o 09 34 06

Ashoat:
Based on screenshots -

1. Web version looks good.
2. First native version breaks the "Preserving simple messages" directive in #37 by adding a padding above and below all messages.
3. It's not clear to me why links are broken in the native version, given that there is an "autolink" rule specified in `react-native-markdown-package`.
4. Where is the background color in "Native version without hyperlink"?

KatPo:
Ad 2 - yeah, I saw that after I compared more carefully, its easy to fix with styling
Ad 3 - for some reason, when Markdown component is inside Hyperlink it returns object and is not displayed, don't know why
Ad 4 - style for message background is currently in Hyperlink, if we remove hyperlink it should be in TouchableOpacity

Ashoat:
Okay awesome! Based on that it seems like you're pretty close!!

KatPo:
> Okay awesome! Based on that it seems like you're pretty close!!

yeah, I guess I just need to reuse this to create our own markdown component and hopefully figure out what's wrong with lists?

Ashoat:
Yeah! Once you've verified that the approach will work (I guess links are all that's left?), you can start on the actual PRs.

I would start by just replacing `react-native-hyperlink` (on native) and `react-linkify` (on web) with a single-rule `autolink` markdown parser, as described by my comment last night in #37.

KatPo:
For now, we have rules for autolink and to preserve simple messages, but since simple-markdown uses rule for link in autolink, I've also included this one.

KatPo:
One more thing - for now, I had to remove styling emoji-only text, need to figure out how to put this in styles for markdown.

native/chat/inner-text-message.react.js
```diff
@@ -78,20 +64,11 @@ class InnerTextMessage extends React.PureComponent<Props> {
...
-        >
-          <Text
-            onPress={this.props.onPress}
-            onLongPress={this.props.onPress}
-            style={[textStyle, textCustomStyle]}
-          >
-            {text}
-          </Text>
-        </Hyperlink>
+        <Markdown onLink={url => Linking.openURL(url)} darkStyle={darkColor}>
```
Ashoat, line 69: Let's move `onLink` to a static method on the class so the function doesn't get reconstructed on every render

native/components/message-markdown-rules.react.js
```diff
@@ -0,0 +1,155 @@
+// @flow
+
+import * as React from 'react';
+import { Text } from 'react-native';
+import SimpleMarkdown from 'simple-markdown';
+
+type State = { [string]: any };
```
Ashoat, line 7: There are a lot of `any` types here. In the case where `simple-markdown` types it as `any`, it's okay to use. But let's try to avoid it as much as possible.

Also - if any of these types are possibly to import directly from `simple-markdown` or a file within it, that would be preferable to redefining them.

native/components/message-markdown-rules.react.js
```diff
@@ -0,0 +1,155 @@
...
+
+type ASTNode = SingleASTNode | Array<SingleASTNode>;
+
+type SingleASTNode = {
+  type: string,
+  [string]: any,
+};
+
+type ReactElement = React$Element<any>;
```
Ashoat, line 16: We should always prefer `React.Element` types to `React$Element` types. Here's a recent comment from a Flow team member in the Flow Discord chat:

> React.Thing is better. React$Thing is the internal version of the type. The qualified types are aliases that may have default type arguments (see AbstractComponent for example)

native/components/message-markdown-rules.react.js
```diff
@@ -0,0 +1,155 @@
...
+
+type Output<Result> = (node: ASTNode, state?: ?State) => Result;
+
+export default function(styles: any, opts: any = {}) {
+  const LINK_INSIDE = '(?:\\[[^\\]]*\\]|[^\\]]|\\](?=[^\\[]*\\]))*';
+  const LINK_HREF_AND_TITLE =
+    '\\s*<?([^\\s]*?)>?(?:\\s+[\'"]([\\s\\S]*?)[\'"])?\\s*';
+  var pressHandler = function(target) {
```
Ashoat, line 24: 1. Let's avoid `var` declarations in favor of `const`, or `let` if necessary
2. I think it's cleaner to use the newer closure syntax:

```js
const pressHandler = target => {
  [...]
};
```

(I'm mentioning this here once, but please make sure to apply it everywhere in the PR!)

native/components/message-markdown-rules.react.js
```diff
@@ -0,0 +1,155 @@
...
+      react: function(
+        node: SingleASTNode,
+        output: Output<ReactElement>,
+        state: State,
+      ) {
+        state.withinText = true;
+        const _pressHandler = () => {
+          pressHandler(node.target);
+        };
+        const autolink = React.createElement(
```
Ashoat, line 45: We should prefer JSX syntax (which gets transpiled to this) since it's more readable.

(I'm mentioning this here once, but please make sure to apply it everywhere in the PR!)

native/components/message-markdown-rules.react.js
```diff
@@ -0,0 +1,155 @@
...
+            key: state.key,
+            style: styles.autolink,
+            onPress: _pressHandler,
+          },
+          output(node.content, state),
+        );
+        return autolink;
+      },
+    },
+    link: {
```
Ashoat, line 57: Do we need `link` for `autolink` to work? If not, let's get rid of it

native/components/message-markdown-rules.react.js
```diff
@@ -0,0 +1,155 @@
...
+            style: styles.autolink,
+            onPress: _pressHandler,
+          },
+          output(node.content, state),
+        );
+        state.withinLink = false;
+        return link;
+      },
+    },
+    paragraph: {
```
Ashoat, line 85: Can you add a code comment explaining why we need `paragraph`?

native/components/message-markdown-rules.react.js
```diff
@@ -0,0 +1,155 @@
...
+          {
+            key: state.key,
+            style: styles.paragraph,
+          },
+          output(node.content, state),
+        );
+        return paragraph;
+      },
+    },
+    text: {
```
Ashoat, line 102: Same here, would be great to have a code comment explaining why we need `text`

native/components/message-markdown-rules.react.js
```diff
@@ -0,0 +1,155 @@
...
+          {
+            key: state.key,
+            style: textStyle,
+          },
+          node.content,
+        );
+        return text;
+      },
+    },
+    url: {
```
Ashoat, line 132: If `url` is just the `<https://google.com>` syntax, let's get rid of it for now. We're only trying to tackle `autolink` in this PR. The goal is for the behavior to be the same as before

native/components/message-markdown.react.js
```diff
@@ -0,0 +1,79 @@
+// @flow
+
+import { Component } from 'react';
```
Ashoat, line 3: I prefer `import * as React from 'react';`, and then you can use `React.Component`

native/components/message-markdown.react.js
```diff
@@ -0,0 +1,79 @@
...
+import * as React from 'react';
+import { View, ViewPropTypes } from 'react-native';
+import { merge, pick } from 'lodash';
```
Ashoat, line 6: 1. Let's use the `lodash/fp` family instead of the older `lodash` one
2. We should never import directly from `lodash` or `lodash/fp` because it bloats bundle size. We can just import the functions we need one-by-one

Check out how I import from `lodash/fp` elsewhere in the project

native/components/message-markdown.react.js
```diff
@@ -0,0 +1,79 @@
...
+import PropTypes from 'prop-types';
+import SimpleMarkdown from 'simple-markdown';
+
+import styles from '../themes/markdown-styles';
+import darkStyles from '../themes/markdown-dark-styles';
+import rules from './message-markdown-rules.react';
+
+type Props = {
```
Ashoat, line 14: We've talked before about avoiding the ambiguous `{ }` syntax. Always prefer exact syntax `{| |}`, or inexact syntax `{...}` if you need it.

Please try to catch these sorts of errors next time before submitting the PR! In my ideal world, I only ever have to give you feedback on something like this once for you to incorporate it into your future work.

A good approach is to always run `git diff master...` before submitting your PR, and imagine that you're me reading it. Try to catch anything I would catch before I do.

native/components/message-markdown.react.js
```diff
@@ -0,0 +1,79 @@
...
+  onLink: (url: string) => Promise<any>,
+  styles: typeof styles,
+  rules: any,
```
Ashoat, line 17: Does this need to be `any`?

native/components/message-markdown.react.js
```diff
@@ -0,0 +1,79 @@
...
+  darkStyle: boolean,
+  children: React.Node,
+};
+class Markdown extends Component<Props> {
```
Ashoat, line 21: Always prefer `React.PureComponent` in this project unless there's an explicit reason otherwise

native/components/message-markdown.react.js
```diff
@@ -0,0 +1,79 @@
...
+  static propTypes = {
+    onLink: PropTypes.func,
```
Ashoat, line 23: All of these `propTypes` are optional, which doesn't match your Flow types

native/components/message-markdown.react.js
```diff
@@ -0,0 +1,79 @@
...
+    darkStyle: PropTypes.bool,
+    children: PropTypes.node,
+  };
+
+  static defaultProps = {
+    styles: styles,
+    rules: {},
+    darkStyle: false,
+  };
+  parse: any;
```
Ashoat, line 37: Does this need to be `any`?

native/components/message-markdown.react.js
```diff
@@ -0,0 +1,79 @@
...
+    super(props);
+    const opts = {
+      onLink: props.onLink,
+      rules: props.rules,
+    };
+
+    const style = props.darkStyle ? darkStyles : styles;
+    var customRules = rules(style, opts);
+
+    var customMarkdownRules = pick(SimpleMarkdown.defaultRules, [
```
Ashoat, line 50: Why do we need these?

native/components/message-markdown.react.js
```diff
@@ -0,0 +1,79 @@
...
+      'autolink',
+      'link',
+      'reflink',
+      'text',
+      'paragraph',
+      'url',
+    ]);
+
+    customRules = merge({}, customMarkdownRules, customRules, opts.rules);
```
Ashoat, line 59: It looks like rules can come from way too many places. Let's simplify this logic. Can we keep all of our rules in `message-markdown-rules.react.js`?

web/package.json
```diff
@@ -60,6 +60,7 @@
     "react-feather": "^2.0.3",
     "react-hot-loader": "^4.12.14",
     "react-linkify": "^1.0.0-alpha",
+    "react-markdown": "^4.3.1",
```
Ashoat, line 63: This package isn't used elsewhere in the PR, I think you forgot to remove it

native/themes/markdown-dark-styles.js
```diff
@@ -0,0 +1,39 @@
+// @flow
+
+import { StyleSheet } from 'react-native';
```
Ashoat, line 3: I don't think we need to have these in separate files. Most of the styles are identical.

Let's figure out a way to have just a single stylesheet that behaves like `styleSelector`. Instead of having two `autolink` styles, you'll just specify one like this:

```js
  autolink: {
    color: 'link',
    textDecorationLine: 'underline',
  },
```

We won't be able to use `styleSelector` in this way exactly, since it decides whether the content is dark based on app-wide dark mode, rather than based on the color of the chat bubble. But we should be able to reuse `stylesFromColors` from `native/themes/colors.js` for it.

native/components/message-markdown-rules.react.js
KatPo: simple-markdown uses `link` for `autolink`, so we need to have it

native/components/message-markdown.react.js
KatPo: Maybe the name is not the best, but basically we need to choose which rules from `SimpleMarkdown.defaultRules` we want to use, if we just use all of them and user types something recognized by `simple-markdown` we'll get an error because we don't have components to match and create in our rules

native/components/message-markdown.react.js
KatPo: I guess we can get rid of `opts.rules`, since we don't need to add more based on component we're creating

web/package.json
KatPo: My bad, it was for the initial commit to show markdown on web version

native/themes/markdown-dark-styles.js
KatPo: I'm aware that they are almost the same, just didn't know exactly how to handle this, but I think your idea will work

native/components/message-markdown-rules.react.js
Ashoat: Ah, I see... looking at the code now. `autolink`, `mailto`, and `url` all go through `link`.

native/components/message-markdown.react.js
Ashoat: I guess I'm confused why we need the `defaultRules` when we've defined the rules in `message-markdown-rules.react.js` already

KatPo: `defaultRules` contain some additional data, like order, that our rules don't

Ashoat: Make sense. In that case, let's define those extra data inline in `message-markdown-rules.react.js`:

```js
{
  autolink: {
    ...SimpleMarkdown.defaultRules.autolink,
    react: function(
      node: SingleASTNode,
      output: Output<ReactElement>,
      state: State,
    ) {
      [...]
    },
  },
}
```

Ashoat:
Let me know when this is ready for a review! (I'm assuming you're still working on it since it's listed as WIP)

native/components/message-markdown-rules.react.js
KatPo: actually, I think simple-markdown uses `url` for matching `https`, `autolink` regex is for `<https://google.com>` syntax, among others

native/components/message-markdown.react.js
KatPo: Will remember for the future, since we don't need `lodash` now with the rules coming from one place

native/components/message-markdown-rules.react.js
```diff
@@ -1,110 +1,102 @@
...
         ),
       ),
-      react: function(
-        node: SingleASTNode,
-        output: Output<ReactElement>,
-        state: State,
-      ) {
+      react: (
+        node: SimpleMarkdown.SingleASTNode,
+        output: SimpleMarkdown.Output<ReactElement>,
```
KatPo, line 63: I think importing like this is making this `any`, so let me know which option would be better to use

Ashoat: I'm confused - importing what is making what `any`?

KatPo: When I don't specify those types, but import from `simple-markdown`, flow is treating them as `any`

Ashoat: Here's how I would've asked this question:

> I am not sure how to import the types from `simple-markdown`. I've noticed they're defined in [this file](https://github.com/Khan/simple-markdown/blob/master/simple-markdown.js), but when I try using them like this:
> ```js
> import SimpleMarkdown from 'simple-markdown';
> let output: SimpleMarkdown.Output<ReactElement>...
> ```
> I see that `output` is getting typed as `any`. Am I doing something wrong here?

Here are some things you could try:
1. `import * as SimpleMarkdown from 'simple-markdown';`
2. `import type { Output } from 'simple-markdown/simple-markdown';`

native/themes/colors.js
```diff
@@ -215,11 +215,16 @@ function useOverlayStyles<IS: Styles>(obj: IS): StyleSheetOf<IS> {
   ]);
 }
 
+function markdownStyle<IS: Styles>(obj: IS, theme: string): StyleSheetOf<IS> {
```
Ashoat, line 218: 1. `theme` should be type `GlobalTheme`
2. Let's call this function `getStylesForTheme`

native/themes/markdown-styles.js
```diff
@@ -0,0 +1,44 @@
...
+    flexDirection: 'row',
+  },
+  view: {
+    alignSelf: 'stretch',
+  },
+};
+
+export type Style = $Exact<typeof unboundStyles>;
+
+export function style(theme: string) {
```
Ashoat, line 42: 1. Let's name this export something more specific. How about `getMarkdownStyles`
2. `theme: GlobalTheme`

native/themes/markdown-styles.js
```diff
@@ -0,0 +1,44 @@
...
+  },
+  textRow: {
+    flexDirection: 'row',
+  },
+  view: {
+    alignSelf: 'stretch',
+  },
+};
+
+export type Style = $Exact<typeof unboundStyles>;
```
Ashoat, line 40: 1. Let's name this export something more specific. How about `MarkdownStyles`
2. Is `typeof unboundStyles` not exact already? You can figure this out by running:

`flow type-at-pos themes/markdown-styles.js 5 7`

native/components/message-markdown.react.js
```diff
@@ -0,0 +1,51 @@
...
+    darkStyle: false,
+  };
+
+  parse: SimpleMarkdown.Parser;
+  renderer: SimpleMarkdown.ReactOutput;
+  style: StyleSheetOf<Style>;
+
+  render() {
+    const theme = this.props.darkStyle ? 'dark' : 'light';
+    this.style = style(theme);
```
Ashoat, line 36: `this.style` is an example of derived data that we don't want to recalculate on every render. We talked over video chat previously about the main rules for when we need to memoize derived data:
1. If the data can be identical without being strictly equal (objects can be like this, but primitives like strings cannot), and
2. If the data is expensive to recalculate. In our case we have to redefine all of this stuff - `parser`, `renderer`, etc. - which is quite expensive

When both of these rules apply, we know we need to figure out a way to memoize the derived data. The main ways to do this memoization are:

1. `getDerivedStateFromProps` - this React lifecycle function lets us recalculate derived data before `render`. With this solution, we would keep all the derived data in `this.state`. We would also cache the old value of `this.props.darkStyle` in `this.state`, and then in `getDerivedStateFromProps` compare `darkStyle` to see if it changes. If it changes, we would recalculate the derived data and return an object to update the state
2. `createSelector` - `reselect` is designed for derived data from Redux. In our case our input isn't coming from Redux, but we could just create a selector that takes a different object as input, and construct a new version of that object on every `render`. It looks like that object would just have one property: `darkStyle`. `reselect` would make sure that it only recalculates the derived data if `darkStyle` changes (this is what "memoization" means)
3. React Hooks - `React.useMemo` can handle this for us. We could define `parse`, `renderer`, and `style` each using `React.useMemo`, and pass it `darkStyle` in the "deps array" to make sure `React.useMemo` only recalculates these three when `darkStyle` changes.

React Hooks are my preferred solution here. `getDerivedStateFromProps` is generally not recommended by the React team unless you're implementing a "transitioner" within a class component, and `createSelector` is better for cases where we data that comes from Redux, and we want an easy way to reuse it that works with both class components and hooks.

Can you try reimplementing this component as a hook, and making sure the derived data is memoized?

native/components/message-markdown-rules.react.js
```diff
@@ -0,0 +1,136 @@
...
+    text: {
+      ...SimpleMarkdown.defaultRules.text,
+      react: (
+        node: SimpleMarkdown.SingleASTNode,
+        output: SimpleMarkdown.Output<string>,
+        state: SimpleMarkdown.State,
+      ) => {
+        let textStyle = {
+          ...styles.text,
+          ...(state.style || {}),
```
Ashoat, line 100: What is `state.style` used for?

native/components/message-markdown-rules.react.js
```diff
@@ -0,0 +1,136 @@
...
+        };
+
+        if (state.withinLink) {
```
Ashoat, line 103: That's interesting, didn't realize you could have a `text` node within a `link` node. I'd like to understand how `SimpleMarkdown`'s parser works better. Can you give me an example of Markdown syntax that will trigger `withinLink` here, and the corresponding AST that `SimpleMarkdown`'s parser generates from it?

native/components/message-markdown-rules.react.js
```diff
@@ -0,0 +1,136 @@
...
+    //simple-markdown uses url rule to catch https inside text
+    url: {
+      ...SimpleMarkdown.defaultRules.url,
+      react: (
+        node: SimpleMarkdown.SingleASTNode,
+        output: SimpleMarkdown.Output<SimpleMarkdown.ReactElement>,
+        state: SimpleMarkdown.State,
+      ) => {
+        state.withinText = true;
+        const _pressHandler = () => {
```
Ashoat, line 124: I am concerned that so much of this code still seems to be copy-pasted. We talked previously about how you shouldn't copy-paste code from other packages. You can look it at side-by-side, and use it as inspiration. But when you name the variables the same and define the functions in the same way, it seems that you are copy-pasting.

This is why I suggested in my previous review that you start over from scratch. The fact that you started this PR by copy-pasting code is an issue. If you really don't want to start over, you need to make sure that you go through all of this code and rewrite every line.

native/components/message-markdown-rules.react.js
```diff
@@ -0,0 +1,136 @@
...
+    // in some cases, simple-markdown recognizes "normal" text as a paragraph
+    // and we need this rule to preserve our style and bubble size
+    paragraph: {
+      ...SimpleMarkdown.defaultRules.paragraph,
+      react: (
+        node: SimpleMarkdown.SingleASTNode,
+        output: SimpleMarkdown.Output<SimpleMarkdown.ReactElement>,
+        state: SimpleMarkdown.State,
+      ) => {
+        const paragraph = (
```
Ashoat, line 82: There's no need to define this variable here, you can just return this directly

native/components/message-markdown-rules.react.js
```diff
@@ -0,0 +1,136 @@
...
+            {output(node.content, state)}
+          </Text>
+        );
+        state.withinLink = false;
+        return link;
+      },
+    },
+    // in some cases, simple-markdown recognizes "normal" text as a paragraph
+    // and we need this rule to preserve our style and bubble size
+    paragraph: {
```
Ashoat, line 75: I'd like to understand how SimpleMarkdown's parser works better. I still don't understand the difference between "paragraph" and "text", or between "autolink"/"link"/"url".

For each of these, can you give me an example of Markdown syntax that will trigger them? I'd ideally like to also see the corresponding AST that SimpleMarkdown's parser generates from it.

native/components/message-markdown-rules.react.js
```diff
@@ -0,0 +1,136 @@
...
+      react: (
+        node: SimpleMarkdown.SingleASTNode,
+        output: SimpleMarkdown.Output<SimpleMarkdown.ReactElement>,
+        state: SimpleMarkdown.State,
+      ) => {
+        state.withinText = true;
+        const _pressHandler = () => {
+          pressHandler(node.target);
+        };
+        const autolink = (
```
Ashoat, line 40: There's no need to define this variable here, you can just return this directly

native/components/message-markdown-rules.react.js
KatPo: Basically, every `link` node contains `text` node inside. Content in `text` is equal to target in `link` 
Zrzut ekranu 2020-07-7 o 11 20 33

native/components/message-markdown-rules.react.js
KatPo: My bad, it was additional style from another rule I was trying to create and forgot to remove

native/components/message-markdown-rules.react.js
KatPo: Paragraph is like a main node, holding other inside it's content. If content is contains something more than just plain text (for example text, then some link ) then it will be splited into separate nodes, with appropriate types (in this case, text and link). Paragraph is responsible for wrapping message bubble over text.
Text is just a normal text, styled like link or plain text.
Autolink matches `<https://google.com>` syntax, url  `https://www.google.com` (or ` https://www.google.com` or other starting with http), and link will be triggered by `[google link](https://www.google.com)`. It's not needed for us, at least now, but we need to have link since both autolink and url are going through link in `simple-markdown`.
Zrzut ekranu 2020-07-7 o 11 52 24
Zrzut ekranu 2020-07-7 o 11 52 59
Zrzut ekranu 2020-07-7 o 11 55 38

native/chat/inner-text-message.react.js
```diff
@@ -111,27 +90,11 @@ class InnerTextMessage extends React.PureComponent<Props> {
 }
 
 const styles = StyleSheet.create({
-  darkLinkText: {
-    color: colors.light.link,
-    textDecorationLine: 'underline',
-  },
-  emojiOnlyText: {
```
Ashoat, line 118: We need to figure out a way to keep the increased-emoji-size logic before we can merge this

native/themes/colors.js
```diff
@@ -179,8 +179,9 @@ function stylesFromColors<IS: Styles>(
 function styleSelector<IS: Styles>(
   obj: IS,
 ): (state: AppState) => StyleSheetOf<IS> {
-  return createSelector(colorsSelector, (themeColors: Colors) =>
-    stylesFromColors(obj, themeColors),
+  return createSelector(
+    colorsSelector,
+    (themeColors: Colors) => stylesFromColors(obj, themeColors),
```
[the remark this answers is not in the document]
KatPo: Yes, but I'm not sure why, now it's back to previous version

KatPo:
I think I've checked all the cases for emoji only message and it looks like before

native/components/message-markdown.react.js
```diff
@@ -0,0 +1,42 @@
...
+  const parser = React.useMemo(() => SimpleMarkdown.parserFor(customRules), [
+    customRules,
+  ]);
+
+  const parse = React.useMemo(() => {
+    // we want to parse childen as one block, instead of inline
+    // to be interpreted as a one block, simple-markdown rules require \n\n at the end
+    // we're adding this manually, and specyfying inline as false, so block-parsing is forced
+    const blockSource = props.children + '\n\n';
+    return parser(blockSource, { inline: false });
```
Ashoat, line 30: Reading through the code of `simple-markdown`, it looks like `inline: false` is the default, and if `disableAutoBlockNewlines: true` is not set, `simple-markdown` will append the newlines for us. So I think the `\n\n` and `{ inline: false }` are not necessary. We could simplify to:

```js
const parse = React.useMemo(
  () => parser(props.children),
  [parser, props.children],
);
```

That said, I think we might want to go with `{ inline: true }` – see my top-level comment about line breaks.

native/components/message-markdown.react.js
```diff
@@ -0,0 +1,42 @@
...
+  const customRules = React.useMemo(() => rules(style), [style]);
+
+  const parser = React.useMemo(() => SimpleMarkdown.parserFor(customRules), [
+    customRules,
+  ]);
+
+  const parse = React.useMemo(() => {
+    // we want to parse childen as one block, instead of inline
+    // to be interpreted as a one block, simple-markdown rules require \n\n at the end
+    // we're adding this manually, and specyfying inline as false, so block-parsing is forced
```
Ashoat, line 28: Thanks for adding all these comments! Two style nits that I'd like you to apply to all of the comments:

1. Let's try to keep line length to a maximum of 80 characters. Prettier doesn't handle enforcing this for code comments, so we have to do it manually
2. This one matters a lot less, but while we're at it let's capitalize the first letter in all of the sentences

Before I merge this PR I'll probably go through all of the comments and reword them a little bit to be clearer.

native/components/message-markdown.react.js
```diff
@@ -0,0 +1,42 @@
...
+import * as SimpleMarkdown from 'simple-markdown';
+
+import { getMarkdownStyles } from '../themes/markdown-styles';
+import rules from './rules.react';
+
+type Props = {|
+  darkStyle: boolean,
+  children: string,
+|};
+function MessageMarkdown(props: Props) {
```
Ashoat, line 14: Let's just call this `Markdown` instead of `MessageMarkdown`, since none of it is specific to messages. I think we'll want to use this same component for calendar entries down the line :)

native/components/message-markdown.react.js
KatPo: If we go with `inline: true`, it will break in some cases, for example this: `[test](https://google.com) example` will be displayed like this
Zrzut ekranu 2020-07-9 o 09 10 52
instead of just 
Zrzut ekranu 2020-07-9 o 09 11 47

native/components/rules.react.js
```diff
@@ -0,0 +1,135 @@
...
+        output: SimpleMarkdown.Output<SimpleMarkdown.ReactElement>,
+        state: SimpleMarkdown.State,
+      ) {
+        return (
+          <Text key={state.key} style={styles.paragraph}>
+            {output(node.content, state)}
+          </Text>
+        );
+      },
+    },
```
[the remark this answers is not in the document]
KatPo: Eslint is complaining about missing display name when I'm trying to do this

native/components/rules.react.js
```diff
@@ -0,0 +1,135 @@
...
+        state: SimpleMarkdown.State,
+      ) {
+        let textStyle = styles.text;
+
+        if (state.linkPresent) {
+          textStyle = styles.link;
+        }
+        if (state.emojiOnly) {
+          textStyle = styles.emojiOnlyText;
+        }
```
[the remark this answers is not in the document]
KatPo: Right, also I've added one more check for emojiOnly, it should cover this ` [👽](https://google.com)`.

native/components/rules.react.js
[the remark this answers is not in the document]
Ashoat: Yeah, ESLint thinks this function is a React component because it's a function that returns a `React.Node`. But in our case we never use it as a `<Component>`, so it's fine. Add an `eslint-ignore` line

native/components/rules.react.js
```diff
@@ -0,0 +1,142 @@
...
+        output: SimpleMarkdown.Output<string>,
+        state: SimpleMarkdown.State,
+      ) {
+        const textStyle = [styles.text];
+        if (state.linkPresent) {
+          textStyle.push(styles.link);
+        }
+        if (state.emojiOnly && !state.linkPresent) {
+          textStyle.push(styles.emojiOnlyText);
+        }
```
Ashoat, line 120: This should be:

```js
if (state.linkPresent) {
  textStyle.push(styles.link);
} else if (state.emojiOnly) {
  textStyle.push(styles.emojiOnlyText);
}
```

native/themes/markdown-styles.js
```diff
@@ -0,0 +1,35 @@
...
+    flexWrap: 'wrap',
+    justifyContent: 'flex-start',
+  },
+  text: {
+    color: 'listForegroundLabel',
+    fontFamily: 'Arial',
+    fontSize: 18,
+  },
+  emojiOnlyText: {
+    fontFamily: 'Arial',
```
Ashoat, line 26: The point of using an array for `textStyle` was so that we didn't have to repeat lines like `fontFamily: 'Arial'`. We only need this in the `text` style. You can just copy-paste the `unboundStyles` array I gave you on my last review

KatPo: Sorry, forgot to add markdown-styles to files to push

native/chat/inner-text-message.react.js
```diff
@@ -61,14 +55,6 @@ class InnerTextMessage extends React.PureComponent<Props> {
     }
     textCustomStyle.height = item.contentHeight;
```
Ashoat, line 56: We're no longer setting this. I'll fix it, but try to catch it next time

Ashoat: Actually, I realized this will be quite complex..

native/chat/inner-text-message.react.js
```diff
@@ -59,15 +58,9 @@ class InnerTextMessage extends React.PureComponent<Props> {
       messageStyle.backgroundColor = this.props.colors.listChatBubble;
       darkColor = this.props.activeTheme === 'dark';
     }
-    textCustomStyle.height = item.contentHeight;
```
Ashoat, line 62: I forgot about an important part here that is going to be pretty complex to solve.

To properly layout our chat list, we need to know the height of every single item. This can be difficult, since the height of a chat message depends on the device width and how the text wraps.

Before this PR, we were using `TextHeightMeasurer` to measure the height of the text blob when rendered with the appropriate device width. We need to update that logic to make sure it's measuring the height of the markdown-rendered components.

This will be pretty challenging, so I'll try to handle it.

native/chat/inner-text-message.react.js
```diff
@@ -78,20 +71,11 @@ class InnerTextMessage extends React.PureComponent<Props> {
...
-        >
-          <Text
-            onPress={this.props.onPress}
-            onLongPress={this.props.onPress}
-            style={[textStyle, textCustomStyle]}
-          >
-            {text}
-          </Text>
-        </Hyperlink>
+        <Text style={[styles.text, textCustomStyle]}>
```
Ashoat, line 76: I brought `styles.text` back here since this allows us to get rid of `state.linkPresent`. We need it before because the style inside the `text` rule was overriding the style from the `link` rule, but if we specify the `text` style at the root, this isn't an issue

native/markdown/rules.react.js
```diff
@@ -0,0 +1,80 @@
...
+        state: SimpleMarkdown.State,
+      ) {
+        if (state.emojiOnly) {
+          return (
+            <Text key={state.key} style={styles.emojiOnlyText}>
+              {node.content}
+            </Text>
+          );
+        }
+        return <React.Fragment key={state.key}>{node.content}</React.Fragment>;
```
Ashoat, line 76: Using `React.Fragment` here because we don't need to specify any style

native/markdown/rules.react.js
```diff
@@ -0,0 +1,80 @@
...
+import type { StyleSheetOf } from '../themes/colors';
+import type { MarkdownStyles } from './styles';
+
+import * as React from 'react';
+import { Text, Linking } from 'react-native';
+import * as SimpleMarkdown from 'simple-markdown';
+
+export default function rules(styles: StyleSheetOf<MarkdownStyles>) {
+  return {
+    // Matches '<https://google.com>' during parse phase and returns a 'link'
```
Ashoat, line 12: I shortened most of your comments. Sorry if it feels like I wasted your time with that. I mostly just wanted it to be clear what each rule was matching and how the AST looked

native/markdown/styles.js
```diff
@@ -0,0 +1,27 @@
+// @flow
```
Ashoat, line 1: I moved all the new files into a new `markdown` directory

Ashoat:
In #104 I noticed yet another issue with newlines. If I sent a message like `Test\n\nA` it looked like `TestA`. I also tested it against the version of this PR before https://github.com/Ashoat/squadcal/pull/67/commits/04d79735e1ffc0f2dba9a8cc0fea520dae4b64ef, and found it was broken in a different way: it looked like `Test\nA`.

I think I've got it fixed with https://github.com/Ashoat/squadcal/pull/67/commits/04d79735e1ffc0f2dba9a8cc0fea520dae4b64ef. Things I had to change:

1. I wasn't able to figure out a way to set the text style in `InnerTextMessage`. Whatever style I set would mess up the children, and collapse newlines together. So instead, I am passing a prop `textStyle` to the `Markdown` component, which then passes it to the logic in `rules.react.js` through the `state` variable.
2. The prop `darkStyle` seemed confusing next to `textStyle`, since one is a style but the other is a `boolean`. So I renamed `darkStyle` to `useDarkStyle`.
3. Because the text style is now set in the rules again, I had to bring `linkPresent` back.
4. I had to edit the `paragraph` RegEx again, this time to make sure it didn't match any trailing newlines. We want every single newline to be converted to its own `paragraph`. This time I simplified the RegEx greatly so it just matches single lines. I also had to switch from `anyScopeRegex` to `blockRegex` so we wouldn't match paragraphs within paragraphs.
5. Because the RegEx change made the `\n\n` that gets appended turn into two `paragraph` elements, I had to disable those with `disableAutoBlockNewlines: true`.
6. I made the emoji-only style just 2x the `fontSize` of the `textStyle` we get passed in.
7. I had to set the `paragraph` style to `nowrap` so every paragraph would start a new line.

Requests for you:

1. Can you rebase #104 against the updated version of this branch?
2. When you were working on this PR, I know you tested a bunch of edge cases. Can you test them again on this updated version to make sure things look good?
3. Can you make sure to test the `web` version of the `Markdown` component to make sure it can handle newlines well? I'm wondering if we'll need to port any of the above changes over to `web`. The important thing to test is that it should never merge newlines together. This should hold true even if there is a newline at the start or end, since we need to calculate an `Entry`'s height while somebody is editing it and adds a newline.

native/markdown/rules.react.js
```diff
@@ -33,12 +44,9 @@ export default function rules(styles: StyleSheetOf<MarkdownStyles>) {
         const onPressLink = () => {
           Linking.openURL(node.target);
         };
-        state.linkPresent = true;
-        const innerNode = output(node.content, state);
-        state.linkPresent = false;
```
Ashoat: Since we're setting the `textStyle` in the parent again, we don't need `linkPresent`

native/markdown/rules.react.js
```diff
@@ -49,61 +57,45 @@ export default function rules(styles: StyleSheetOf<MarkdownStyles>) {
       ...SimpleMarkdown.defaultRules.paragraph,
       // simple-markdown collapses multiple newlines into one, but we want to
       // preserve the newlines
-      match: SimpleMarkdown.blockRegex(/^([^\n]*)(?:\n|$)/),
+      match: SimpleMarkdown.blockRegex(/^((?:[^\n]*)(?:\n|$))/),
```
Ashoat, line 60: Note that the capture group (the outer parentheses) now includes the newline character

native/markdown/rules.react.js
```diff
@@ -49,61 +57,45 @@ export default function rules(styles: StyleSheetOf<MarkdownStyles>) {
...
         output: SimpleMarkdown.Output<SimpleMarkdown.ReactElement>,
         state: SimpleMarkdown.State,
-      ) => {
-        let innerNode = output(node.content, state);
-        if (Array.isArray(innerNode) && innerNode.length === 0) {
-          // React Native renders multiple empty lines within their own Text
-          // differently than it renders a single Text that contains those empty
-          // lines within it. We render this space to make sure it renders as if
-          // it were a single Text node
-          innerNode = ' ';
```
Ashoat: We don't need this anymore because we're not rendering block-level elements

native/markdown/rules.react.js
```diff
@@ -49,61 +57,45 @@ export default function rules(styles: StyleSheetOf<MarkdownStyles>) {
...
-        }
-        return (
-          <Text key={state.key} style={styles.paragraph}>
```
Ashoat: We don't need the `paragraph` style anymore, so we can just return a `React.Fragment`. So clean!

native/markdown/rules.react.js
```diff
@@ -49,61 +57,45 @@ export default function rules(styles: StyleSheetOf<MarkdownStyles>) {
...
-          style.push({ fontSize: fontSize * 2 });
-        }
-
-        return (
-          <Text key={state.key} style={style}>
-            {node.content}
-          </Text>
-        );
-      },
+      ) => <React.Fragment key={state.key}>{node.content}</React.Fragment>,
```
Ashoat, line 81: Even cleaner! The `Markdown` component handles `emojiOnly` now, and the `link` rule handles itself

native/markdown/rules.react.js
```diff
@@ -49,61 +57,45 @@ export default function rules(styles: StyleSheetOf<MarkdownStyles>) {
...
   };
+  return {
+    simpleMarkdownRules,
+    emojiOnlyFactor: null,
+  };
+}
+
+// We allow the most markdown features for TextMessage, which doesn't have the
+// same requirements as Entry
+function fullMarkdownRules(
```
Ashoat, line 92: I renamed these and moved the `export` declaration to the bottom

native/markdown/rules.react.js
```diff
@@ -6,12 +6,23 @@ import type { MarkdownStyles } from './styles';
...
 import * as SimpleMarkdown from 'simple-markdown';
-import invariant from 'invariant';
 
 import { urlRegex } from 'lib/shared/markdown';
 
-export default function rules(styles: StyleSheetOf<MarkdownStyles>) {
-  return {
+type MarkdownRuleSpec = {|
+  +simpleMarkdownRules: SimpleMarkdown.ParserRules,
+  +emojiOnlyFactor: ?number,
```
Ashoat, line 14: I wanted to make the `emojiOnly` decision part of the "rules", so I created a new structure that include both `simple-markdown` rules and `emojiOnlyFactor`

native/markdown/markdown.react.js
```diff
@@ -1,47 +1,79 @@
...
+      flattened && typeof flattened === 'object',
+      `state passed to Markdown output should have textStyle`,
+    );
+    const { fontSize } = flattened;
+    invariant(
+      fontSize,
+      `textStyle should have fontSize if using emojiOnlyFactor`,
+    );
+    return { ...flattened, fontSize: fontSize * emojiOnlyFactor };
+  }, [emojiOnly, props.textStyle, emojiOnlyFactor]);
```
Ashoat, line 74: We determine the `textStyle` here based on `emojiOnly`

native/markdown/markdown.react.js
```diff
@@ -1,47 +1,79 @@
...
   );
 
-  const output = React.useMemo(() => {
-    return SimpleMarkdown.outputFor(customRules, 'react');
-  }, [customRules]);
-  const emojiOnly = React.useMemo(() => onlyEmojiRegex.test(props.children), [
-    props.children,
-  ]);
-  const { textStyle } = props;
-  return React.useMemo(() => output(ast, { emojiOnly, textStyle }), [
```
Ashoat: We no longer have to pass any state in here

PR #82, opened 2020-06-22
[native] Add number of unread background threads to title

KatPo:


KatPo:
If I remember correctly, unread background threads should not be counted in the badge, but they are
Zrzut ekranu 2020-06-22 o 11 48 18

native/chat/chat-thread-background.react.js
```diff
@@ -15,6 +20,35 @@ type BackgroundChatThreadListProps = {
 export default function BackgroundChatThreadList(
   props: BackgroundChatThreadListProps,
 ) {
+  const chatThreadListData = useSelector(state => chatListData(state));
+
+  const findUnreadBackgroundThreadsNumber = (
```
Ashoat, line 25: Why define this function separately, instead of just inline in the `useMemo` below?

(Based on my comment below we'll want to keep it separate so we can use it in `web`. But in this initial implementation where we were defining it inline in `BackgroundChatThreadList`, I think it would've made sense to just put it inside the `useMemo`.)

native/chat/chat-thread-background.react.js
```diff
@@ -15,6 +20,35 @@ type BackgroundChatThreadListProps = {
...
+      .filter(item => !item.home && item.unread).length;
+  };
+
+  const prevUnreadNumber = React.useRef(0);
+  const unreadBackgroundThreadsNumber = React.useMemo(() => {
+    return findUnreadBackgroundThreadsNumber(chatThreadListData);
+  }, [chatThreadListData]);
+
+  React.useEffect(() => {
+    if (unreadBackgroundThreadsNumber == prevUnreadNumber.current) return;
```
Ashoat, line 42: Some nits:

1. Let's always default to using the `===` comparison
2. Let's always include curly braces for blocks, and put the contents on a new line:

```js
if (unreadBackgroundThreadsNumber === prevUnreadNumber.current) {
  return;
}
```

native/chat/chat-thread-background.react.js
```diff
@@ -15,6 +20,35 @@ type BackgroundChatThreadListProps = {
...
+    prevUnreadNumber.current = unreadBackgroundThreadsNumber;
+    let title = 'Background';
+    if (unreadBackgroundThreadsNumber !== 0)
+      title = title.concat(' (', unreadBackgroundThreadsNumber.toString(), ')');
```
Ashoat, line 46:
```js
if (unreadBackgroundThreadsNumber !== 0) {
  title += ` (${unreadBackgroundThreadsNumber})`;
}
```

native/chat/chat-thread-background.react.js
```diff
@@ -15,6 +20,35 @@ type BackgroundChatThreadListProps = {
...
+    props.navigation.setOptions({
+      title: title,
```
Ashoat, line 48: Modern JavaScript let's you turn `{ title: title }` into just `{ title }`:

```js
props.navigation.setOptions({ title });
```

native/chat/chat-thread-background.react.js
```diff
@@ -15,6 +20,35 @@ type BackgroundChatThreadListProps = {
...
+    });
+  });
```
Ashoat, line 50: Right now you've set this effect to run every time the component renders. Let's add a dep list here so it only updates when it needs to.

native/chat/chat-thread-background.react.js
```diff
@@ -15,6 +20,35 @@ type BackgroundChatThreadListProps = {
 export default function BackgroundChatThreadList(
   props: BackgroundChatThreadListProps,
 ) {
+  const chatThreadListData = useSelector(state => chatListData(state));
+
+  const findUnreadBackgroundThreadsNumber = (
+    chatThreadsListData: Array<ChatThreadItem>,
+  ): number => {
+    return chatThreadsListData
```
Ashoat, line 28: Let's define this function in `lib`, since we anticipate using it on the `web` side pretty soon.

We can put it next to `unreadCount` in `lib/selectors/thread-selectors.js`, since it's a similar problem. And let's have it depend directly on `state.threadStore.threadInfos` there instead of going through `chatListData`.

native/chat/chat-thread-background.react.js
KatPo: Check if the dep lists in `useMemo` and `useEffect` are correct now, with selector used to get unread number

lib/selectors/thread-selectors.js
```diff
@@ -173,6 +173,19 @@ const unreadCount: (state: BaseAppState<*>) => number = createSelector(
...
 
+const unreadBackgroundCount: (
+  state: BaseAppState<*>,
+) => number = createSelector(
+  (state: BaseAppState<*>) => state.threadStore.threadInfos,
+  (threadInfos: { [id: string]: RawThreadInfo }): number =>
+    _flow(
+      _filter(threadInChatList),
+      _filter(thread => !thread.currentUser.subscription.home),
+      _filter('currentUser.unread'),
```
zrebcu411, line 184: Nit: I prefer not to use this kind of string shorthands. Flow would never catch it if `RawThreadInfo` type would change in the future

KatPo: I did this like it was in `unreadCount`, but I also prefer not to use them

KatPo: Actually, `thread.currentUser.subscription.home` was supposed to be done the same way, but for now it's reversed value so I did it traditionally

Ashoat: Good call @zrebcu411. I have a lot of Lodash usage across this project that I would prefer to decrease. In particular these string shorthands can't be typechecked.

I also noticed that the first two conditions here are basically `threadInBackgroundChatList`. What if we updated this function to look like this:

```js
import { values } from '../utils/objects';

import { threadInBackgroundChatList } from '../shared/thread-utils';

const unreadBackgroundCount: (
  state: BaseAppState<*>,
) => number = createSelector(
  (state: BaseAppState<*>) => state.threadStore.threadInfos,
  (threadInfos: { [id: string]: RawThreadInfo }): number => values(threadInfos).filter(
    threadInfo => threadInBackgroundChatList(threadInfo) && threadInfo.currentUser.unread,
  ).length,
);
```

native/chat/chat-thread-background.react.js
```diff
@@ -15,6 +17,26 @@ type BackgroundChatThreadListProps = {
 export default function BackgroundChatThreadList(
   props: BackgroundChatThreadListProps,
 ) {
+  const unreadBackgroundThreadsCount = useSelector(state =>
+    unreadBackgroundCount(state),
+  );
+
+  const prevUnreadNumber = React.useRef(0);
+  const unreadBackgroundThreadsNumber = React.useMemo(() => {
```
Ashoat, line 25: This call to `useMemo` doesn't do anything

native/chat/chat-thread-background.react.js
```diff
@@ -15,6 +17,26 @@ type BackgroundChatThreadListProps = {
...
+    return unreadBackgroundThreadsCount;
+  }, [unreadBackgroundThreadsCount]);
+
+  React.useEffect(() => {
+    if (unreadBackgroundThreadsNumber === prevUnreadNumber.current) {
+      return;
+    }
+    prevUnreadNumber.current = unreadBackgroundThreadsNumber;
+    let title = 'Background';
+    if (unreadBackgroundThreadsNumber !== 0)
```
Ashoat, line 35: As mentioned in the previous review, please add the curly braces

lib/selectors/thread-selectors.js
```diff
@@ -168,7 +170,7 @@ const unreadCount: (state: BaseAppState<*>) => number = createSelector(
   (threadInfos: { [id: string]: RawThreadInfo }): number =>
     _flow(
       _filter(threadInChatList),
-      _filter('currentUser.unread'),
+      _filter(threadInfo => threadInfo.currentUser.unread),
```
Ashoat, line 173: Nice! Thanks for fixing this up

PR #84, opened 2020-06-23
[native] Add message for empty background threads tab

KatPo:


KatPo:
I've tried to create stacked pr, not sure if I did it correctly though

Ashoat:
Yeah, the stacked PR didn't go through correctly. Stacked PRs are rather complicated in GitHub, and require you to wait for all of the stack to be approved, and then you have to merge backwards (from the new ones to the old ones).

[Here's a link](https://unhashable.com/stacked-pull-requests-keeping-github-diffs-small/) that explains how to do it. If you want to try this for next time, let me know so that I don't merge the old PRs first (I've already merged #82).

For now - since I've already merged #82, you can just rebase against master and it should look good.

native/chat/chat-thread-list.react.js
```diff
@@ -124,6 +129,18 @@ class ChatThreadList extends React.PureComponent<Props, State> {
         />
       );
     }
+    if (item.type === 'empty') {
```
Ashoat, line 132: Let's avoid having code here that is specific to background threads. Background thread stuff should live in `BackgroundChatThreadList`.

How about making `emptyItem` a `React.Node`, and passing this node in from `BackgroundChatThreadList`?

KatPo: To clarify: do you mean only the `Text` component with  explanation?

KatPo:
We might want to do the same for home threads, potentially user can move all of his threads to background.

native/chat/chat-thread-background.react.js
```diff
@@ -34,10 +36,38 @@ export default function BackgroundChatThreadList(
     props.navigation.setOptions({ title });
   }, [props.navigation, unreadBackgroundThreadsNumber]);
 
+  const textStyle =
```
Ashoat, line 39: Instead of deciding on the colors in the component, let's try to use the colors defined in the themes in `native/themes/colors.js`. This way we can use the same colors everywhere, and if we want to change any colors, we only need to edit one place.

In this case, since it looks like `ChatThreadList` has `backgroundColor: 'listBackground'`, we can use `color: 'listBackgroundLabel'` for our text. This is similar to what you've selected, only it uses `#C7C7CC` instead of `white`. (Let me know if it makes the text too hard to see, I haven't actually checked what it looks like. An alternative would be `listForegroundLabel`, which is exactly the same colors you're using here.)

To see how we can use these colors in a Hook, check out `native/chat/new-messages-pill.react.js`. You basically define the style sheet pretending that `listBackgroundLabel` is a color, and then you pass it to `useStyles`, which will decide if it's a light or dark background.

Note that you don't need to call `StyleSheet.create` in this case. `useStyles` will call it for you, and I think in `DEV` mode `StyleSheet.create` can error if you give it a color it doesn't understand.

native/chat/chat-thread-list.react.js
Ashoat: Yeah, that's what I meant!

Let's use `item.emptyItem` here instead of `this.props.emptyItem`. In this case it actually makes no difference since they are the same, but as a rule of thumb `renderItem` shouldn't be looking at `this.props` because when `this.props` changes, there's no guarantee `renderItem` will be run again. `renderItem` only gets run again when the `item` objects change.

This is why we always try to put the props into the item. (In case where a single prop affects every item, an alternate approach is to pass that prop in as `extraData` to the `FlatList`.)

native/chat/chat-thread-list.react.js
```diff
@@ -181,6 +191,13 @@ class ChatThreadList extends React.PureComponent<Props, State> {
           searchResults.has(item.threadInfo.id),
         );
       }
+      if (this.props.emptyItem && chatItems.length === 0) {
+        const emptyItem = this.props.emptyItem;
```
Ashoat, line 195: Nit:

```js
const { emptyItem } = this.props;
```

native/chat/chat-thread-list.react.js
```diff
@@ -43,13 +43,17 @@ const floatingActions = [
...
+  | ChatThreadItem
+  | {| type: 'search', searchText: string |}
+  | {| type: 'empty', emptyItem: React.Node |};
 
 type Props = {|
   navigation: ChatTopTabsNavigationProp<
     'HomeChatThreadList' | 'BackgroundChatThreadList',
   >,
   filterThreads: (threadItem: ThreadInfo) => boolean,
+  emptyItem: React.Node,
```
Ashoat, line 56: Since this is an optional property, let's specify it as:

```js
  emptyItem?: React.Node,
```

(Flow doesn't catch this because the type of our `connect` function is broken, and I have never got around to fixing it...)

native/chat/chat-thread-list.react.js
KatPo: I see one problem with using `item.emptyItem` here: when user changes theme, from light to dark for example, `listDataSelector` is not called again, `item.emptyItem` is not changing, and therefore until `listDataSelector` is called again, we're still displaying the old `item.emptyItem`.

KatPo:
I didn't see any problems on `FlatList` loading, it looks fine every time

native/chat/chat-thread-background.react.js
```diff
@@ -34,10 +36,31 @@ export default function BackgroundChatThreadList(
...
   return (
     <ChatThreadList
       navigation={props.navigation}
       filterThreads={threadInBackgroundChatList}
+      emptyItem={emptyItem}
     />
   );
 }
+
+const style = {
```
Ashoat, line 59: Let's call this `styles` for consistency with other places

native/chat/chat-thread-list.react.js
Ashoat: Great catch!! You're totally right, this is an issue here. The reason it's happening is because we're checking `this.props.emptyItem` directly in `listDataSelector`, instead of selecting it from `propsAndState`. I should've noticed this on the first review.

One concern about selecting it from `propsAndState` is that since we're generating a new `React.Node` in `ChatThreadBackground` every time it renders, that means we will end up generating a new `data` for the `FlatList` in `ChatThreadList` every time `ChatThreadBackground` renders.

Here's an idea - instead of passing in a `React.Node`, let's pass in a React component (`React.ComponentType`). We can define a component for our `emptyItem` that calls `useStyles`, and pass that component in from `ChatThreadBackground` to `ChatThreadList`. Then that component will handle re-rendering itself when dark mode changes, without us having to rerender the entire `FlatList`.

native/chat/chat-thread-background.react.js
KatPo: Since we can't have `const styles = useStyles(styles);`, one of them must be inconsistent ( in `native/chat/new-messages-pill.react.js` we have `const styles = useStyles(unboundStyles);` )

Ashoat: Right, good catch. Let's call it `unboundStyles` then

KatPo:
It's working like we want it to work, however I'm not sure if I did this correctly (especially if props are typed correctly), so let me know if I need to change something.

native/chat/chat-thread-background.react.js
```diff
@@ -34,10 +36,34 @@ export default function BackgroundChatThreadList(
     props.navigation.setOptions({ title });
   }, [props.navigation, unreadBackgroundThreadsNumber]);
 
+  const emptyItem = <EmptyItem />;
```
Ashoat, line 39: The issue I raised from the last review is still happening:

> we're generating a new `React.Node` in `ChatThreadBackground` every time it renders, that means we will end up generating a new data for the `FlatList` in `ChatThreadList` every time `ChatThreadBackground` renders

Here's the solution I proposed in the last review:

> instead of passing in a `React.Node`, let's pass in a React component (`React.ComponentType`)

If this doesn't make sense, try Googling those two and seeing what the difference is. If it's still confusing, feel free to message me or @zrebcu411 and we can help explain.

native/chat/chat-thread-list.react.js
```diff
@@ -181,6 +197,13 @@ class ChatThreadList extends React.PureComponent<Props, State> {
           searchResults.has(item.threadInfo.id),
         );
       }
+      if (this.props.emptyItem && chatItems.length === 0) {
```
Ashoat, line 200: Let's get `emptyItem` from `propsAndState`. It's working now when you switch dark mode only because `ChatThreadBackground` never changes `emptyItem`. But calling `this.props` directly here means that if `ChatThreadBackground` changes `emptyItem`, the change won't get picked up by `ChatThreadList`. That's a bad way to build things - a component should never assume that a prop it's being passed won't change.

I tried to cover this in my last review:

> The reason it's happening is because we're checking `this.props.emptyItem` directly in `listDataSelector`, instead of selecting it from `propsAndState`

native/chat/chat-thread-background.react.js
```diff
@@ -38,6 +40,29 @@ export default function BackgroundChatThreadList(
     <ChatThreadList
       navigation={props.navigation}
       filterThreads={threadInBackgroundChatList}
+      emptyItem={EmptyItem}
     />
   );
 }
+
+const EmptyItem: React.ComponentType<{||}> = () => {
```
Ashoat, line 48: Let's just do:

```js
function EmptyItem() {
```

native/chat/chat-thread-list.react.js
```diff
@@ -124,6 +129,9 @@ class ChatThreadList extends React.PureComponent<Props, State> {
         />
       );
     }
+    if (item.type === 'empty') {
+      return <item.emptyItem />;
```
Ashoat, line 133: Just a style nit here:

```js
const EmptyItem = item.emptyItem;
return <EmptyItem />;
```

native/chat/chat-thread-list.react.js
```diff
@@ -166,10 +182,12 @@ class ChatThreadList extends React.PureComponent<Props, State> {
     (propsAndState: PropsAndState) => propsAndState.chatListData,
     (propsAndState: PropsAndState) => propsAndState.searchText,
     (propsAndState: PropsAndState) => propsAndState.searchResults,
+    (propsAndState: PropsAndState) => propsAndState.emptyItem,
     (
       reduxChatListData: $ReadOnlyArray<ChatThreadItem>,
       searchText: string,
       searchResults: Set<string>,
+      emptyItem?: any,
```
Ashoat, line 190: Fix the type here

native/chat/chat-thread-list.react.js
```diff
@@ -181,6 +199,12 @@ class ChatThreadList extends React.PureComponent<Props, State> {
           searchResults.has(item.threadInfo.id),
         );
       }
+      if (emptyItem && chatItems.length === 0) {
```
Ashoat, line 202: How about:

```js
if (emptyItem && chatItems.length === 0) {
  chatItems.push({ type: 'empty', emptyItem });
}
```

This way we avoid defining `{ type: 'search', searchText }` twice

PR #87, opened 2020-06-24
[native] Rename"Home notifs" to "Background" and invert the value

KatPo:


KatPo:
The last two commits should be on a different branch, sorry about that. Let me know if I should close this pr and open a new one with just them.

Ashoat:
This looks good, but yeah we need to separate the changes out.

PR #89, opened 2020-06-24
[server] Add other update relationship actions

zrebcu411:


server/src/updaters/relationship-updaters.js
```diff
@@ -40,7 +41,11 @@ async function updateRelationships(
     }
   }
 
-  if (action === 'send_friend_request') {
+  if (!userIDs.length) {
+    return errors;
+  }
+
+  if (relationshipActions.FRIEND) {
```
Ashoat, line 48: You forgot the `action ===` here. Would be awesome if we could catch small errors like this before review. Making sure you test the endpoint will help a lot, as will *carefully* going through every line in your PR before putting it up.

server/src/updaters/relationship-updaters.js
```diff
@@ -91,14 +96,64 @@ async function updateRelationships(
...
+    `;
+
+    await Promise.all([dbQuery(insertQuery), dbQuery(deleteQuery)]);
+
+    return errors;
+  } else if (action === relationshipActions.BLOCK) {
+    const directedRows = [];
+    const undirectedRows = [];
+    for (const userID of userIDs) {
+      const ids = [userID, viewer.userID].map(Number).sort((a, b) => a - b);
```
Ashoat, line 129: Nit: maybe move `const ids` declaration down one line, since it's only used for `undirectedRows`

server/src/updaters/relationship-updaters.js
```diff
@@ -91,14 +96,64 @@ async function updateRelationships(
...
     if (directedDeleteQuery.length) {
       promises.push(dbQuery(directedDeleteQuery));
     }
 
     await Promise.all(promises);
 
     return { ...errors, ...friendRequestErrors };
+  } else if (action === relationshipActions.UNFRIEND) {
+    const updateRows = userIDs.map(userID => {
+      const ids = [userID, viewer.userID].map(Number).sort((a, b) => a - b);
```
Ashoat, line 108: Maybe we can extract this into a top-level function, since we use it a lot?

server/src/updaters/relationship-updaters.js
```diff
@@ -91,14 +96,64 @@ async function updateRelationships(
...
+    const directedQuery = SQL`
+      INSERT INTO relationships_directed (user1, user2, status)
+      VALUES ${directedRows}
+      ON DUPLICATE KEY UPDATE status = VALUES(status)
+    `;
+    const undirectedQuery = SQL`
+      INSERT INTO relationships_undirected (user1, user2, status)
+      VALUES ${undirectedRows}
+      ON DUPLICATE KEY UPDATE status = VALUES(status)
+    `;
```
Ashoat, line 143: If user2 has friend requested user1, should we delete that friend request here?

server/src/updaters/relationship-updaters.js
```diff
@@ -91,14 +96,64 @@ async function updateRelationships(
...
+    const insertQuery = SQL`
+      INSERT INTO relationships_undirected (user1, user2, status)
+      VALUES ${updateRows}
+      ON DUPLICATE KEY UPDATE status = VALUES(status)
+    `;
+    const deleteQuery = SQL`
+      DELETE FROM relationships_directed
+      WHERE status = ${directedStatus.PENDING_FRIEND} AND 
+        user1 = ${viewer.userID} AND user2 IN (${userIDs})
+    `;
```
Ashoat, line 120: I think we'll use this endpoint for rejecting somebody's friend request too, which means we should delete friend requests in either direction

server/src/updaters/relationship-updaters.js
zrebcu411: Yes, definitely

server/src/updaters/relationship-updaters.js
zrebcu411: I was wondering about that. If we will do that it would be possible to decline friend request but the requester could invite again and again. But I guess that's how it works on social apps. If a user doesn't want to receive friend requests anymore he can block the annoying requester. I've updated a query to delete in both directions then

PR #95, opened 2020-06-30
[server] Create relationship user updates

zrebcu411:


.vscode/launch.json
```diff
@@ -0,0 +1,27 @@
+{
+  // Use IntelliSense to learn about possible attributes.
+  // Hover to view descriptions of existing attributes.
+  // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+  "version": "0.2.0",
+  "configurations": [
```
Ashoat: I'm curious, what does this file do?

lib/types/user-types.js
```diff
@@ -2,9 +2,18 @@
...
+  | 'request_received'
+  | 'blocker'
+  | 'blocked'
+  | 'both_blocked'
+  | 'friends'
+  | 'know_of';
 export type UserInfo = {|
   id: string,
   username: ?string,
+  relationshipStatus?: RelationshipStatus,
```
Ashoat: I think we need to update the following as well:

- `userInfoPropType`
- `AccountUserInfo`
- `accountUserInfoPropType`
- `RelativeUserInfo`
- `relativeUserInfoPropType`

server/src/updaters/relationship-updaters.js
```diff
@@ -102,8 +108,10 @@ async function updateRelationships(
 
     await Promise.all(promises);
 
-    return { ...errors, ...friendRequestErrors };
+    errors = { ...errors, ...friendRequestErrors };
```
Ashoat, line 111: Nit: let's move this line up to where `friendRequestErrors` is defined

PR #98, opened 2020-07-01
[web] Add background threads support

KatPo:


KatPo:
Two things: right now active thread might be in background threads, and since we're not checking that and default tab is Home, we might need to take this into consideration, but I wasn't sure what to do with this. 
This pr doesn't have the number of unread background threads (Background (1) ) yet

web/chat/chat-tabs.react.js
```diff
@@ -0,0 +1,58 @@
+// @flow
+
+import * as React from 'react';
+import PropTypes from 'prop-types';
+
+import ChatThreadTab from './chat-thread-tab.react';
+import css from './chat-tabs.css';
+
+type Props = {|
+  children: Array<React.Element<any>>,
```
zrebcu411, line 10: Maybe make it more specific: `children: Array<React.Element<typeof ThreadsTab>>`?

web/chat/threads-tab.react.js
```diff
@@ -0,0 +1,11 @@
+// @flow
+
+import * as React from 'react';
+
+type Props = {
+  title: string,
+  ...
```
zrebcu411, line 7: Can we make it exact? `{| title: string |}`

web/chat/threads-tab.react.js
```diff
@@ -0,0 +1,11 @@
...
+};
+export default function ThreadsTab(props: Props) {
+  return <div title={props.title} />;
```
Ashoat, line 10: I don't think we render this

web/chat/threads-tab.react.js
Ashoat: I don't think we need this component at all, but if we were keeping it we'd want to list all of its props here. I think you're just missing `children` (which is maybe why you got errors when you tried to make it exact?)

web/chat/chat-thread-list.react.js
```diff
@@ -26,14 +29,25 @@ type Props = {|
...
 
   render() {
-    const threads = this.props.chatListData.map(item => (
+    const chatListData = this.props.chatListData.filter(item =>
+      this.props.filterThreads(item.threadInfo),
+    );
+
+    if (chatListData.length === 0 && this.props.emptyItem) {
+      const EmptyItem = this.props.emptyItem;
+      return <div className={css.container}>{<EmptyItem />}</div>;
```
Ashoat, line 47: The curly braces around `<EmpyItem />` aren't necessary

Ashoat: We can avoid specifying `<div className={css.container}>` twice here:

```js
const threads: React.Node[] = this.props.chatListData.filter(
  item => this.props.filterThreads(item.threadInfo),
).map(item => (
  <ChatThreadListItem
    item={item}
    active={item.threadInfo.id === this.props.navInfo.activeChatThreadID}
    navInfo={this.props.navInfo}
    timeZone={this.props.timeZone}
    dispatchActionPayload={this.props.dispatchActionPayload}
    key={item.threadInfo.id}
  />
));
if (threads.length === 0 && this.props.emptyItem) {
  const EmptyItem = this.props.emptyItem;
  threads.push(<EmptyItem />);
}
return <div className={css.container}>{threads}</div>;
```

web/chat/chat-tabs.react.js
```diff
@@ -0,0 +1,58 @@
...
+|};
+type State = {|
+  activeTab: string,
+|};
+
+class ChatTabs extends React.PureComponent<Props, State> {
```
Ashoat, line 16: You call this `ChatThreadTabs` elsewhere. Let's use consistent names for components - the export and import names for component should be the same, and the filename should be the same as the component name

web/chat/threads-tab.react.js
KatPo: yes, it was because of that, but since I wasn't using `children`, I wasn't listing them

web/chat/threads-tab.react.js
KatPo: yes, now with the approach you proposed, I've removed ThreadsTab since we don't need this

web/chat/chat-tabs.react.js
KatPo: my bad, renamed the file and component, but left import names, sorry

web/chat/chat-thread-list.css
```diff
@@ -1,11 +1,10 @@
 div.container {
-  position: absolute;
+  position: relative;
   width: 400px;
   top: 0;
   bottom: 0;
   overflow-y: auto;
   background-color: white;
-  border-right: 2px solid #E4E4E4;
```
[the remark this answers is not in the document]
KatPo: you're right, it's not needed now

web/chat/chat-tabs.react.js
```diff
@@ -0,0 +1,56 @@
...
+  state = {
+    activeTab: 'HOME',
+  };
+
+  onClickTabItem = (tab: string) => {
+    this.setState({ activeTab: tab });
+  };
+  homeTab = (
```
Ashoat, line 23: It seems that `homeTab` and `backgroundTab` are initialized when the class is constructed, and don't get redefined if `this.state.activeTab` changes. Does the underline actually move when you press a button?

I don't think they should be class fields, we can probably just define them in `render`. That also would allow us to skip the `key` prop.

```js
    return (
      <div className={css.container}>
        <ol>
          <ChatThreadTab
            activeTab={this.state.activeTab}
            title="HOME"
            onClick={this.onClickTabItem}
          />
          <ChatThreadTab
            activeTab={this.state.activeTab}
            title="BACKGROUND"
            onClick={this.onClickTabItem}
          />
        </ol>
        <div>{threadList}</div>
      </div>
    );
```

web/chat/chat-tabs.react.js
```diff
@@ -0,0 +1,56 @@
...
+    const threadTabs = [this.homeTab, this.backgroundTab];
+    const threadList =
+      this.state.activeTab === 'HOME' ? (
+        <ChatThreadHome />
+      ) : (
+        <ChatThreadBackground />
+      );
+    return (
+      <div className={css.container}>
+        <ol>{threadTabs}</ol>
```
Ashoat, line 49: I don't think it's correct HTML to use a `<ol>` without `<li>` inside

web/chat/chat-tabs.react.js
```diff
@@ -0,0 +1,56 @@
...
+    <ChatThreadTab
+      activeTab={this.state.activeTab}
+      key="BACKGROUND"
+      title="BACKGROUND"
+      onClick={this.onClickTabItem}
+    />
+  );
+  render() {
+    const threadTabs = [this.homeTab, this.backgroundTab];
+    const threadList =
```
Ashoat, line 41: Nit: I think this would look cleaner:

```js
const { activeTab } = this.state;
const threadList =
  activeTab === 'HOME' ? <ChatThreadHome /> : <ChatThreadBackground />;
```

web/chat/chat.react.js
```diff
@@ -3,13 +3,14 @@
...
 
-import ChatThreadList from './chat-thread-list.react';
 import ChatMessageList from './chat-message-list.react';
 import InputStateContainer from '../input/input-state-container.react';
+import ChatTabs from './chat-tabs.react';
 
 type Props = {|
   setModal: (modal: ?React.Node) => void,
 |};
+
```
Ashoat, line 13: Nit: I've noticed you like to add newlines here. I'd prefer to leave this class as-is

web/chat/chat-thread-list.react.js
```diff
@@ -26,24 +28,34 @@ type Props = {|
 |};
 class ChatThreadList extends React.PureComponent<Props> {
   static propTypes = {
+    filterThreads: PropTypes.func.isRequired,
+    emptyItem: PropTypes.element,
```
Ashoat, line 32: This should be a `PropTypes.elementType` (corresponds to `React.ComponentType`), not a `PropTypes.element` (corresponds to `React.Element`)

web/chat/chat-tabs.css
```diff
@@ -0,0 +1,43 @@
...
+  position: absolute;
+  width: 400px;
+  top: 0;
+  bottom: 0;
+  overflow-y: auto;
+  background-color: white;
+  border-right: 2px solid #E4E4E4;
+}
+
+.tabList {
```
Ashoat, line 11: We don't seem to be using this?

web/chat/chat-tabs.css
```diff
@@ -0,0 +1,43 @@
...
+  border-bottom: 1px solid #ccc;
+  padding-left: 0;
+  align-items: center;
+  justify-content: space-between;
+  background-color: white;
+}
+  
+.tabItem {
```
Ashoat, line 19: When defining a CSS selector, let's always specify the element type (unless we need to use the selector for multiple element types): `button.tabItem`, `button.tabItemActive`

web/chat/chat-thread-background.react.js
```diff
@@ -0,0 +1,29 @@
...
+      filterThreads={threadInBackgroundChatList}
+      emptyItem={EmptyItem}
+    />
+  );
+}
+
+function EmptyItem() {
+  return (
+    <div className={css.emptyItem}>
+      Background threads are just like normal threads, except they appear in
```
Ashoat, line 22: Instead of copy-pasting this text, let's put it in `lib/shared/thread-utils.js`, and import it from both native and web. That way if we can change it in one place we'll change it in the other

web/chat/chat-tabs.react.js
KatPo: yeah, you're totally right, underline is not moving. At first I defined them in `render`, then changed do class fields, but somehow missed that pressing the button is not changing underline. Sorry about that!

web/chat/chat-tabs.css
```diff
@@ -0,0 +1,35 @@
...
+  top: 0;
+  bottom: 0;
+  overflow-y: auto;
+  background-color: white;
+  border-right: 2px solid #E4E4E4;
+}
+
+button.tabItem {
+  color: grey;
+  width: 200px;
```
Ashoat, line 13: Let's figure out a way to do this without explicitly specifying width here

web/chat/chat-tabs.css
```diff
@@ -0,0 +1,35 @@
...
+  display: inline-block;
+  list-style: none;
```
Ashoat, line 15: Get rid of `list-style`

Ashoat:
I went ahead and updated the PR with some changes. @KatPo - please make sure to review my changes!

1. Get rid of `<button>` usage. I didn't like how it behaved on mouseover and on press. It looks cleaner to just use a basic div
2. I avoided repeating style between `tabItem` and `tabItemActive` by using `classNames`
3. I rebased on master to address the Flow errors. I assume you've been sidestepping the Flow error with `--no-verify`, which is an anti-pattern - let's avoid this in the future

PR #99, opened 2020-07-02
Update prop types

KatPo:


server/package.json
```diff
@@ -49,6 +49,7 @@
     "mysql2": "^1.5.1",
     "node-schedule": "^1.3.0",
     "nodemailer": "^6.3.0",
+    "prop-types": "^15.7.2",
```
Ashoat, line 52: We don't need to add this here

PR #100, opened 2020-07-03
[web] Add unread threads number to background tab

KatPo:


web/chat/chat-tabs.react.js
```diff
@@ -1,18 +1,28 @@
...
+import { unreadBackgroundCount } from 'lib/selectors/thread-selectors';
 
 import ChatThreadTab from './chat-thread-tab.react';
 import css from './chat-tabs.css';
 import ChatThreadHome from './chat-thread-home.react';
 import ChatThreadBackground from './chat-thread-background.react';
 
-type Props = {||};
+type Props = {|
+  unreadBgrCount?: number,
```
Ashoat, line 17: 1. Let's call this `unreadBackgroundCount`
2. `key?: type` means that the property may or may not be specified, but in your case it is always specified, just might be `undefined` sometimes. Instead, let's use `key: ?type`.

The difference is that `key?: type` matches both `{ }` and `{ key: undefined }`, but `key: ?type` only matches the latter.

web/chat/chat-tabs.react.js
```diff
@@ -1,43 +1,60 @@
...
+  };
   state = {
     activeTab: 'HOME',
   };
 
-  onClickTabItem = (tab: string) => {
-    this.setState({ activeTab: tab });
+  onClickHome = () => {
+    this.setState({ activeTab: 'HOME' });
+  };
```
Ashoat, line 32: Nit: let's put a space between these two

web/chat/chat-tabs.react.js
```diff
@@ -1,43 +1,60 @@
...
+  onClickBackground = () => {
+    this.setState({ activeTab: 'BACKGROUND' });
   };
 
   render() {
     const { activeTab } = this.state;
     const threadList =
       activeTab === 'HOME' ? <ChatThreadHome /> : <ChatThreadBackground />;
+    let backgroundTitle = `BACKGROUND`;
```
Ashoat, line 41: Nit: there's no need to use a template string here, just use a normal one

PR #101, opened 2020-07-03
[server] Create know_of relationships on thread join

zrebcu411:


server/src/creators/thread-creator.js
```diff
@@ -86,13 +90,18 @@ async function createThread(
...
+    knowOfRelationshipsChangeset,
   ] = await Promise.all([
     changeRole(id, [viewer.userID], newRoles.creator.id),
     initialMemberIDs && initialMemberIDs.length > 0
       ? changeRole(id, initialMemberIDs, null)
       : undefined,
     recalculateAllPermissions(id, threadType),
+    initialMemberIDs && initialMemberIDs.length > 0
+      ? knowOfRelationshipsForThread(initialMemberIDs.concat(viewer.userID))
+      : undefined,
```
Ashoat, line 102: There is some complexity here that I realized I never explained. Ultimately I think we will need to handle `knowOfRelationshipsForThread` inside `thread-permission-updaters`.

Here's the issue:
1. Let's imagine Bob is trying to create a new child thread. The parent thread has users Bob, Alice, and Jane. For the child thread, Bob adds Alice, but also a new user Joe.
2. Because Joe is now in a child thread that has access to the parent thread, the server will create a `membership` row for the parent thread, and pass that parent thread down to Joe. Joe's `role` for this `membership` row in the DB will be `0`, which is represented as `null` in the `MemberInfo` type in JS. This value indicates Joe is not truly a member of the parent thread, and it's similar in concept to `know_of` - the server needs to make sure Joe has an up-to-date version of the parent thread, but Joe is not technically a member.
3. Since Joe has access to the parent thread, we now need to create a `know_of` row between Joe and Jane. But we won't know that we need this row unless we look at the parent thread's members. Furthermore, since threads can be infinitely nested, this problem requires recursively looking up all the parents until we get to the root.

`thread-permission-updaters` has to handle a very similar set of problems, and ends up recursing to the root thread anyways when it's creating a new thread. I also think it would be a good place to put this `knowOfRelationshipsForThread` logic because all membership changes have to go through there anyways. I'm not 100% sure, but I think you could probably avoid having to make any changes to `thread-creator` or `thread-updaters` if you handle everything in `thread-permission-updaters`.

server/src/updaters/relationship-updaters.js
```diff
@@ -199,7 +201,65 @@ async function updateRelationships(
...
+
+  const rows = changeset.map(row => [row.user1, row.user2, row.status]);
+  const query = SQL`
+    INSERT INTO relationships_undirected (user1, user2, status)
+    VALUES ${rows}
+  `;
+  if (override) {
+    query.append('ON DUPLICATE KEY UPDATE status = VALUES(status)');
+  } else {
+    query.append('ON DUPLICATE KEY UPDATE status = status');
```
Ashoat, line 256: Right now we're only using this with `override = false`. I'm wondering - what's the use case you foresee for `override = true`?

server/src/updaters/relationship-updaters.js
```diff
@@ -199,7 +201,65 @@ async function updateRelationships(
...
+    return;
+  }
+
+  const rows = changeset.map(row => [row.user1, row.user2, row.status]);
+  const query = SQL`
+    INSERT INTO relationships_undirected (user1, user2, status)
+    VALUES ${rows}
+  `;
+  if (override) {
+    query.append('ON DUPLICATE KEY UPDATE status = VALUES(status)');
```
Ashoat, line 254: I think it's always a good idea to use `GREATEST` here when setting `know_of` rows, just in case there is a race between a friend request and a new thread being created, for instance. The only time we ever want to replace a `friend` row with a `know_of` row is when the relationship changes (unfriend or block).

Ashoat, line 254: Nit: let's use SQL template literals here for both of the `query.append` calls. It's true we don't need them here because we don't have any inline variables, but I think it's a good practice to always use the SQL template literals. It makes it clear to the reader that you're writing SQL, and it makes sure that if somebody later adds an inline variable, it will be escaped properly.

server/src/updaters/thread-permission-updaters.js
```diff
@@ -69,6 +69,7 @@ async function changeRole(
     dbQuery(membershipQuery),
     changeRoleThreadQuery(threadID, role),
   ]);
+
```
Ashoat, line 72: Nit: let's avoid making purely whitespace changes to files in a PR. The issue is that it pollutes version control history. We want it to be easy for somebody to [look at a file's history](https://github.com/Ashoat/squadcal/commits/master/server/src/updaters/thread-permission-updaters.js) and see when it's changed.

server/src/scripts/create-relationships.js
```diff
@@ -0,0 +1,51 @@
...
+    endScript();
+  } catch (e) {
+    endScript();
+    console.warn(e);
+  }
+}
+
+async function createKnowOfRelationships() {
+  const [result] = await dbQuery(SQL`
+    SELECT thread, user FROM memberships ORDER BY user
```
Ashoat, line 32: Nit: I always forget which direction is the default for `ORDER BY`, so I like to state it explicitly: `ORDER BY user ASC`

server/src/updaters/relationship-updaters.js
```diff
@@ -199,7 +201,65 @@ async function updateRelationships(
...
+  const combinations = combinations2(newMemberIDs).map(
+    ([firstId, secondId]) => {
+      const [user1, user2] = sortIDs(firstId, secondId);
+      return { user1, user2, status: undirectedStatus.KNOW_OF };
+    },
+  );
+
+  if (threadID) {
+    const query = SQL`
+      SELECT user FROM memberships WHERE thread = ${threadID} ORDER BY user
```
Ashoat, line 220: Nit: I always forget which direction is the default for `ORDER BY`, so I like to state it explicitly: `ORDER BY user ASC`

server/src/updaters/relationship-updaters.js
```diff
@@ -199,7 +201,65 @@ async function updateRelationships(
...
+        );
+      }, [])
+      .concat(combinations);
+  } else {
+    return combinations;
+  }
+}
+
+async function updateUndirectedRelationships(
+  changeset: RelationshipChangeset,
```
Ashoat, line 241: Nit: can we make this take `UndirectedRelationshipRow[]`?

server/src/updaters/relationship-updaters.js
```diff
@@ -199,7 +201,65 @@ async function updateRelationships(
...
 function sortIDs(firstId: string, secondId: string) {
-  return [firstId, secondId].map(Number).sort((a, b) => a - b);
+  return [Number(firstId), Number(secondId)].sort((a, b) => a - b);
 }
 
-export { updateRelationships };
+async function knowOfRelationshipsForThread(
+  newMemberIDs: string[],
+  threadID?: string,
+): Promise<RelationshipChangeset> {
```
Ashoat, line 210: Nit: can we make this return `UndirectedRelationshipRow[]`?

server/src/updaters/relationship-updaters.js
zrebcu411: Good point. Now I'm using `greatest: boolean` flag and reusing that function in `relationship-updaters.js`

server/src/updaters/thread-updaters.js
```diff
@@ -427,30 +429,65 @@ async function updateThread(
...
+      relationshipRows: recalculateRelationshipRows,
+    } = recalculatePermissionsChangeset;
+    membershipRows.push(
+      ...recalculateMembershipRows.filter(
         rowToSave => !newMemberIDs.includes(rowToSave.userID),
       ),
     );
+    const parentMemberIDs = recalculateMembershipRows
+      .filter(rowToSave => !newMemberIDs.includes(rowToSave.userID))
+      .map(rowToSave => rowToSave.userID);
```
Ashoat, line 456: Nit: I would flip the order of this `filter`/`map`, since we only care about `userID` anyways

server/src/updaters/thread-updaters.js
```diff
@@ -427,30 +429,65 @@ async function updateThread(
...
+    const newUserIDs = newMemberIDs.filter(
+      memberID =>
+        !recalculateMembershipRows.find(
+          rowToSave => memberID === rowToSave.userID,
+        ),
+    );
```
Ashoat, line 462: I think there is one more edge case that can break this, but the good news is that it should be easy to fix!

## Changing thread parent while simultaneously adding user that was in old parent

1. Thread A exists with users Beth and Joe
2. Thread B exists with users Bob and Joe
3. Thread C exists with single user Joe and is an open child of Thread A
4. Joe calls `updateThread` on Thread C with two changes: add Beth, and also move it to be an open child of Thread B
5. `recalculateAllPermissions` returns an `{ operation: 'delete', user: 'Beth', thread: 'C' }`, because it sees that Beth used to have permissions for Thread C, but doesn't anymore
6. We need to create a `know_of` between Beth and Bob. But Beth will get filtered out of `newUserIDs` because of the aforementioned `{ operation: 'delete' }`

## Fix

I think we can just update this code to check for `operation` as well. I'd also like to check `threadID` - it doesn't make a difference, but it makes it more clear what we're doing: figuring out what users are new to `request.threadID`.

```js
const newUserIDs = newMemberIDs.filter(
  memberID =>
    !recalculateMembershipRows.find(
      rowToSave => rowToSave.userID === memberID &&
        rowToSave.threadID === request.threadID &&
        rowToSave.operation !== 'delete',
    ),
);
```

server/src/updaters/thread-updaters.js
```diff
@@ -427,30 +429,65 @@ async function updateThread(
...
+      newUserIDs,
+    ).map(pair => {
+      const [user1, user2] = sortIDs(...pair);
+      const status = undirectedStatus.KNOW_OF;
+      return { user1, user2, status };
+    });
+    relationshipRows.push(
+      ...recalculateRelationshipRows,
+      ...parentRelationshipRows,
+    );
```
Ashoat, line 474: This code is very similar to what we do in `createThread`. Can you factor it out so we can share it? I'm thinking of a function `getParentThreadMembershipRowsForNewUsers` that takes `recalculateMembershipRows` and `newMemberIDs`

server/src/updaters/thread-updaters.js
zrebcu411: Do you think it applies also to `createThread`? If we want to reuse this logic in `getParentThreadMembershipRowsForNewUsers` it should 🤔

native/android/gradlew.bat
```diff
@@ -1,100 +1,100 @@
...
+:fail
+rem Set variable GRADLE_EXIT_CONSOLE if you need the _script_ return code instead of
+rem the _cmd.exe /c_ return code!
+if  not "" == "%GRADLE_EXIT_CONSOLE%" exit 1
+exit /b 1
+
+:mainEnd
+if "%OS%"=="Windows_NT" endlocal
+
+:omega
```
zrebcu411, line 100: I have no idea why git sees it as changed... I changed this file to LF instead of CRLF because that was annoying when `git status` showed it as a changed file

Ashoat: I saw something like this earlier, couldn't get Git to reset it. I ended up running some Git command that cleared out my `server/facts` and `server/secrets` directories, would not recommend haha. 

My guess is that we'll need to do something with `.gitattributes`

Ashoat:
> Do you think it applies also to `createThread`? If we want to reuse this logic in `getParentThreadMembershipRowsForNewUsers` it should 🤔

Yeah, I think this logic is fine for `createThread`

native/android/gradlew.bat
zrebcu411: Can this changes stay in this PR? Without it I have problems with switching branches, stashing, etc. Even `--hard` reset doesn't let me get rid of this git changes

zrebcu411: All files we have in repo are LF, only this one is CRLF

Ashoat: It's CRLF on purpose (it's a Windows batch file).

Let's figure out a way to fix this where it doesn't cause issues for you, but also doesn't change the line endings. Have you looked into `.gitattributes`? I'm eating lunch right now, but can take a look in an hour or so

zrebcu411: Yes, I tried to add `.gitattributes` with `*.bat text eol=crlf` and `*.bat binary`. It didn't help. I just restored that `gradlew.bat` file back to CRLF

server/src/updaters/thread-permission-updaters.js
```diff
@@ -282,16 +308,24 @@ async function updateDescendantPermissions(
...
             role,
           });
+
+          for (const [childUserID] of userInfos) {
+            if (childUserID !== userID) {
+              const [user1, user2] = sortIDs(childUserID, userID);
+              const status = undirectedStatus.KNOW_OF;
+              relationshipRows.push({ user1, user2, status });
+            }
+          }
```
Ashoat, line 326: Sorry for not noticing this earlier, but I think we can safely gate this behind an `if (!oldPermissions) {` condition

server/src/updaters/thread-permission-updaters.js
```diff
@@ -594,9 +665,38 @@ function setJoinsToUnread(
...
+  const newUserIDs = newMemberIDs.filter(
+    memberID =>
+      !recalculateMembershipRows.find(
+        rowToSave =>
+          rowToSave.userID === memberID &&
+          rowToSave.threadID === threadID &&
+          rowToSave.operation !== 'delete',
+      ),
+  );
+  const parentRelationshipRows = cartesianProduct(
```
Ashoat, line 685: Nit: we can just `return` this directly now, since it's defined in a function

Ashoat:
Perfect!! Thanks for sticking through this, even when we had to change the approach a couple times.

PR #102, opened 2020-07-09
[web] Add custom markdown with simple rules for web

KatPo:


web/chat/markdown.react.js
```diff
@@ -0,0 +1,33 @@
...
+    children: PropTypes.string.isRequired,
+  };
+
+  ast: SimpleMarkdown.Parser;
+  output: SimpleMarkdown.ReactOutput;
+
+  constructor(props: Props) {
+    super(props);
+
+    const parser = SimpleMarkdown.parserFor(rules());
```
Ashoat, line 23: Can we avoid calling `rules()` twice?

web/chat/rules.react.js
```diff
@@ -0,0 +1,19 @@
...
+  return {
+    // We are using default simple-markdown rules
+    // For more details, look at native/components/rules.react
+    autolink: SimpleMarkdown.defaultRules.autolink,
+    link: SimpleMarkdown.defaultRules.link,
+    text: SimpleMarkdown.defaultRules.text,
+    url: {
+      ...SimpleMarkdown.defaultRules.url,
+      match: SimpleMarkdown.inlineRegex(
+        /^(https?:\/\/[^\s<]+[^<.,:;"')\]\s])/i,
```
Ashoat, line 15: Can we extract this to a shared place in `lib` so we don't specify it twice?

web/chat/rules.react.js
```diff
@@ -0,0 +1,19 @@
+// @flow
+
+import * as SimpleMarkdown from 'simple-markdown';
+
+export default function rules() {
+  return {
+    // We are using default simple-markdown rules
+    // For more details, look at native/components/rules.react
+    autolink: SimpleMarkdown.defaultRules.autolink,
+    link: SimpleMarkdown.defaultRules.link,
```
Ashoat, line 10: We should make sure links open in a new tab. I think we'll need to customize the `render` function of `link` to match the old `linkifyComponentDecorator`

web/chat/markdown.react.js
```diff
@@ -0,0 +1,33 @@
+// @flow
```
Ashoat, line 1: Let's put all the new files in a new `web/markdown` directory

web/chat/markdown.react.js
```diff
@@ -0,0 +1,33 @@
...
+  ast: SimpleMarkdown.Parser;
+  output: SimpleMarkdown.ReactOutput;
+
+  constructor(props: Props) {
+    super(props);
+
+    const parser = SimpleMarkdown.parserFor(rules());
+    this.ast = parser(this.props.children, { inline: true });
+    this.output = SimpleMarkdown.outputFor(rules(), 'react');
+  }
```
Ashoat, line 26: I wonder if we should handle the emoji logic here, like we're doing on native

KatPo: I think we shouldn't change this on web, but since we got rid of state.linkPresent, we might also do the same with state.emojiOnly, and handle this like before in `InnerTextMessage` on native, like before.

PR #104, opened 2020-07-14
[native] Replace hyperlink with custom markdown

KatPo:


native/chat/robotext-message.react.js
```diff
@@ -105,7 +105,15 @@ class RobotextMessage extends React.PureComponent<Props> {
         continue;
       }
       if (splitPart.charAt(0) !== '<') {
-        textParts.push(decodeURI(splitPart));
+        textParts.push(
+          <Markdown
+            key={robotextParts.indexOf(splitPart)}
+            darkStyle={true}
```
KatPo, line 111: Two things I'm not sure about: I've added key, since without it there was a yellow warning. I'm also passing `darkStyle={true}`, because we don't have any info about globalTheme here to pass proper value, but it's just a blue link on light/dark background and on both it looks good.
Let me know if there is a better solution about any of this!

native/calendar/entry.react.js
```diff
@@ -392,19 +391,13 @@ class InternalEntry extends React.Component<Props, State> {
             iosActiveOpacity={opacity}
           >
             <View>
-              <View style={heightStyle} />
-              <Hyperlink
-                linkDefault={true}
-                linkStyle={linkStyle}
-                style={this.props.styles.textContainer}
+              <View style={[heightStyle, this.props.styles.textContainer]} />
```
Ashoat, line 394: Why did you add `this.props.styles.textContainer` to the style here?

native/chat/robotext-message.react.js
Ashoat: Yeah, you need a key here, since you're pushing a React element onto an array. Strings are their own keys, but React elements need explicit keys specified.

1. You need to get the theme data and set `darkStyle` based on that
2. Let's make sure that the key doesn't accidentally collide with the `id` keys used below. `id` keys are numeric, so we can solve this by prefixing a string like `text${robotextParts.indexOf(splitPart)}`
3. Let's avoid calling `indexOf` here, there's no need to search through the string for an index. We can initialize a `let keyIndex = 0;` variable at the start, and just use `text${keyIndex++}` as the key

native/markdown/markdown.react.js
```diff
@@ -6,17 +6,21 @@ import * as SimpleMarkdown from 'simple-markdown';
 import { onlyEmojiRegex } from 'lib/shared/emojis';
 
 import { getMarkdownStyles } from './styles';
-import rules from './rules.react';
+import { rules, advancedRules } from './rules.react';
 
 type Props = {|
   darkStyle: boolean,
   children: string,
+  onlyBasicRules?: boolean,
```
Ashoat, line 14: Let's pass the `rules` function in here instead of having this boolean

native/calendar/entry.react.js
KatPo: To make this look like before, without `this.props.styles.textContainer` it was bigger and looked like this:
Zrzut ekranu 2020-07-14 o 14 02 15
instead of 
Zrzut ekranu 2020-07-14 o 14 02 23

Ashoat: That does seem like an issue, but I'm concerned that you are overriding `heightStyle` with `this.props.styles.textContainer` here. I'll look into it...

Ashoat: So what you're doing here is basically making `heightStyle` useless. `textContainer` is `position: absolute`, which means this `View`'s `height` will only apply to itself.

It's important that the height of this `FlatList` item matches `heightStyle`. Otherwise our ability to calculate the position of elements in the `FlatList` could be broken.

I got it working this way:

```js
<View style={heightStyle} />
<View style={this.props.styles.textContainer}>
  <Text
    style={[this.props.styles.text, textStyle]}
    onLayout={this.onTextLayout}
  >
    <Markdown darkStyle={darkColor}>{rawText}</Markdown>
  </Text>
</View>
```

In the future, remember that it's not just about making things look correct. When you change a line of code, make sure you understand 100% why it was there before, and what it was doing.

native/calendar/entry.react.js
```diff
@@ -392,19 +391,13 @@ class InternalEntry extends React.Component<Props, State> {
...
+                onLayout={this.onTextLayout}
               >
-                <Text
-                  style={[this.props.styles.text, textStyle]}
-                  onLayout={this.onTextLayout}
-                >
-                  {rawText}
-                </Text>
-              </Hyperlink>
+                <Markdown darkStyle={darkColor}>{rawText}</Markdown>
```
Ashoat, line 399: Another issue I noticed in testing. If I add two line breaks after some text, the height calculation on `Markdown` gets broken. This indicates something wrong within the `Markdown` component itself

Ashoat: Hmm, this issue seems to repro for messages. If I send a message `Test\n\nA` it turns into `TestA`. I think we tested this, so I suspect one of our later changes broke it. I'll try to investigate this as well...

Ashoat: I finished the investigation and updated #67. Check out my comment there, and make sure to rebase this PR against the updated `feature/message_formatting` branch!

native/calendar/entry.react.js
```diff
@@ -390,13 +390,15 @@ class InternalEntry extends React.Component<Props, State> {
...
-            <View>
-              <View style={[heightStyle, this.props.styles.textContainer]} />
+            <View style={heightStyle} />
+            <View style={this.props.styles.textContainer}>
               <Text
                 style={[this.props.styles.text, textStyle]}
                 onLayout={this.onTextLayout}
               >
-                <Markdown darkStyle={darkColor}>{rawText}</Markdown>
+                <Markdown textStyle={textStyle} useDarkStyle={darkColor}>
```
KatPo, line 399: This style is not the right one to use, it's temporary just to make this commit

native/ios/Podfile.lock
```diff
@@ -286,8 +286,6 @@ PODS:
   - react-native-ffmpeg/min-lts (0.4.4):
     - mobile-ffmpeg-min (= 4.3.1.LTS)
     - React
-  - react-native-in-app-message (1.0.2):
```
Ashoat, line 289: This is strange - why is `react-native-in-app-message` removed here?

native/package.json
```diff
@@ -61,7 +61,6 @@
     "react-native-floating-action": "^1.19.1",
     "react-native-fs": "2.15.2",
     "react-native-gesture-handler": "^1.6.1",
-    "react-native-hyperlink": "git+https://git@github.com/ashoat/react-native-hyperlink.git#both",
```
Ashoat, line 64: One more thing I realized - we should `git rm patches/react-native-hyperlink+0.0.12.patch`

native/ios/Podfile.lock
KatPo: I was also confused, but it happened after I did `pushd native && yarn remove react-native-hyperlink && popd && yarn cleaninstall`

KatPo: Some weird bug I guess, I had to run `yarn clean install` couple of times and it's back now

Ashoat:
Awesome!! Let's merge this into `feature/message_formatting`

PR #106, opened 2020-07-20
[web] Add more rules for markdown

KatPo:


KatPo:
This is very close to being complete, one rule that is not exactly working is `heading`, but I'm trying to solve this

KatPo:
Also, rules are missing description, I'll add them later

web/markdown/rules.react.js
```diff
@@ -47,8 +71,105 @@ function linkRules(): MarkdownRuleSpec {
...
+        <code key={state.key} className={css.inlineCode}>
+          {node.content}
+        </code>
+      ),
+    },
+    em: SimpleMarkdown.defaultRules.em,
+    strong: SimpleMarkdown.defaultRules.strong,
+    del: SimpleMarkdown.defaultRules.del,
+    underline: {
+      order: SimpleMarkdown.defaultRules.em.order - 0.5,
```
Ashoat, line 126: Why use a different order than the `em` rule? `simple-markdown` uses the same order for `u` and `em`

web/markdown/rules.react.js
```diff
@@ -47,8 +71,105 @@ function linkRules(): MarkdownRuleSpec {
...
+      match: SimpleMarkdown.inlineRegex(/^__([\s\S]+?)__(?!_)/),
```
Ashoat, line 127: What is the purpose of making the "one or more" quantifier lazy here (`+?`)? I can't figure out what the difference would be with a simple "one or more" quantifier (`+`), since you're using a negative lookahead (`(?!_)`) to prevent following underscores.

web/markdown/rules.react.js
```diff
@@ -47,8 +71,105 @@ function linkRules(): MarkdownRuleSpec {
...
+      parse: (
+        capture: SimpleMarkdown.Capture,
+        parse: SimpleMarkdown.Parser,
+        state: SimpleMarkdown.state,
+      ) => ({
+        content: parse(capture[1], state),
+      }),
```
Ashoat, line 134: What's the reason for avoiding `parseCaptureInline`? As long as we don't want to match block elements within, I think the default (from `SimpleMarkdown.defaultRules.u`) makes sense here

web/markdown/rules.react.js
```diff
@@ -47,8 +71,105 @@ function linkRules(): MarkdownRuleSpec {
...
+      // eslint-disable-next-line react/display-name
+      react: (
+        node: SimpleMarkdown.SingleASTNode,
+        output: SimpleMarkdown.Output<SimpleMarkdown.ReactElement>,
+        state: SimpleMarkdown.State,
+      ) => <u key={state.key}>{output(node.content, state)}</u>,
```
Ashoat, line 140: It seems like we could just use the default from `SimpleMarkdown.defaultRules.u` here, since we don't need to specify a CSS class

web/markdown/rules.react.js
```diff
@@ -47,8 +71,105 @@ function linkRules(): MarkdownRuleSpec {
...
+    // heading: {
+    //   ...SimpleMarkdown.defaultRules.heading,
+    //   match: SimpleMarkdown.inlineRegex(/^ *(#{1,6} )([^\n]+?) *(?:\n+|$)/),
+    // },
+    mailto: SimpleMarkdown.defaultRules.mailto,
+    codeBlock: {
+      ...SimpleMarkdown.defaultRules.codeBlock,
+      match: SimpleMarkdown.blockRegex(/^(?: {4}[^\n]+\n*)+(?:\n *)/),
+      parse(capture: SimpleMarkdown.Capture) {
+        var content = capture[0].replace(/^ {4}/gm, '');
```
Ashoat, line 151: Please don't ever copy-paste code or use the `var` keyword!

web/markdown/rules.react.js
```diff
@@ -47,8 +71,105 @@ function linkRules(): MarkdownRuleSpec {
...
+      ) => <u key={state.key}>{output(node.content, state)}</u>,
+    },
+    // heading: {
+    //   ...SimpleMarkdown.defaultRules.heading,
+    //   match: SimpleMarkdown.inlineRegex(/^ *(#{1,6} )([^\n]+?) *(?:\n+|$)/),
+    // },
+    mailto: SimpleMarkdown.defaultRules.mailto,
+    codeBlock: {
+      ...SimpleMarkdown.defaultRules.codeBlock,
+      match: SimpleMarkdown.blockRegex(/^(?: {4}[^\n]+\n*)+(?:\n *)/),
```
Ashoat, line 149: 1. Why are we matching any number of newlines following the code block? (`\n*`)
2. Why are we matching a new line following by any number of spaces at the end? `(?:\n *)`

web/markdown/rules.react.js
```diff
@@ -47,8 +71,105 @@ function linkRules(): MarkdownRuleSpec {
...
+      ) => (
+        <code key={state.key} className={css.inlineCode}>
+          {node.content}
+        </code>
+      ),
+    },
+    em: SimpleMarkdown.defaultRules.em,
+    strong: SimpleMarkdown.defaultRules.strong,
+    del: SimpleMarkdown.defaultRules.del,
+    underline: {
```
KatPo, line 125: Looks like this is not necessary, SimpleMarkdown has `u` rule, I've missed that and added custom one

web/markdown/rules.react.js
```diff
@@ -47,8 +71,105 @@ function linkRules(): MarkdownRuleSpec {
...
+      ...SimpleMarkdown.defaultRules.fence,
+      match: SimpleMarkdown.blockRegex(/`{3}(?:(.*$)\n)?([\s\S]*)`{3}/m),
+      parse(capture: SimpleMarkdown.Capture) {
+        return {
+          type: 'codeBlock',
+          lang: capture[1] || undefined,
+          content: capture[2],
+        };
+      },
+    },
```
Ashoat, line 168: We can avoid the `return` keyword and get rid of an indent:

```js
parse: (capture: SimpleMarkdown.Capture) => ({
  type: 'codeBlock',
  lang: capture[1] || undefined,
  content: capture[2],
}),
```

web/markdown/rules.react.js
```diff
@@ -47,8 +71,105 @@ function linkRules(): MarkdownRuleSpec {
...
+    fence: {
+      ...SimpleMarkdown.defaultRules.fence,
+      match: SimpleMarkdown.blockRegex(/`{3}(?:(.*$)\n)?([\s\S]*)`{3}/m),
+      parse(capture: SimpleMarkdown.Capture) {
+        return {
+          type: 'codeBlock',
+          lang: capture[1] || undefined,
```
Ashoat, line 164: By specifying `lang` here you're going to make `codeBlock`'s `react` method add a `className`. Since we don't actually have CSS classes for custom languages, let's avoid specifying this.

I think we can just use `parseBlock` if we make the `language` capture group in the RegEx a non-capture (`(?:)`) group.

web/markdown/rules.react.js
```diff
@@ -47,8 +71,105 @@ function linkRules(): MarkdownRuleSpec {
...
+        var content = capture[0].replace(/^ {4}/gm, '');
+        return {
+          lang: undefined,
+          content: content,
+        };
+      },
+    },
+    fence: {
+      ...SimpleMarkdown.defaultRules.fence,
+      match: SimpleMarkdown.blockRegex(/`{3}(?:(.*$)\n)?([\s\S]*)`{3}/m),
```
Ashoat, line 160: Nice job with this RegEx! You significantly simplified the code. Some changes I think might make sense:

1. Do you need the `m` modifier at the end? I think it only modifies the behavior of beginning (`^`) and end (`$`) characters, which you aren't using.
2. Can we also match the tilde character? We can use `\1` to make sure we don't match ~~~```:

```js
/(`{3,}|~{3,})(?:(.*$)\n)?([\s\S]*)\1/
```

Since we're now adding a capture group you'll need to update the `parse` function

web/markdown/rules.react.js
```diff
@@ -34,7 +35,30 @@ function linkRules(): MarkdownRuleSpec {
         </a>
       ),
     },
-    text: SimpleMarkdown.defaultRules.text,
+    paragraph: {
+      ...SimpleMarkdown.defaultRules.paragraph,
+      match: SimpleMarkdown.blockRegex(/^((?:[^\n]*)(?:\n|$))/),
```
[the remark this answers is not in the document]
KatPo: yes, since for most of the rules we're gonna need custom regex it'll be needed

web/markdown/rules.react.js
```diff
@@ -47,8 +71,105 @@ function linkRules(): MarkdownRuleSpec {
...
+    ...linkMarkdownRules.simpleMarkdownRules,
+    autolink: SimpleMarkdown.defaultRules.autolink,
+    link: {
+      ...linkMarkdownRules.simpleMarkdownRules.link,
+      match: SimpleMarkdown.defaultRules.link.match,
+    },
+    blockQuote: {
+      ...SimpleMarkdown.defaultRules.blockQuote,
+      // match end of blockQuote by either \n\n or end of string
+      match: SimpleMarkdown.blockRegex(/^( *>[^\n]+(\n[^\n]+)*\n*)+(\n{2,}|$)/),
```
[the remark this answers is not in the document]
KatPo: But in case of `blockQuote`, isn't two newlines at the end best way to catch end of quote? If we want blockQuote and normal text after in one message for example, but also we want to have option for blockQuote to be multiline

web/markdown/rules.react.js
```diff
@@ -47,8 +71,105 @@ function linkRules(): MarkdownRuleSpec {
...
+        capture: SimpleMarkdown.Capture,
+        parse: SimpleMarkdown.Parser,
+        state: SimpleMarkdown.State,
+      ) {
+        var content = capture[0].replace(/^ *> ?/gm, '');
+        return {
+          // remove new line from captured string
+          content: parse(content.trim(), state),
+        };
+      },
```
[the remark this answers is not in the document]
KatPo: I've added `trim` to remove captured empty line if we have normal text after `blockQuote`

web/markdown/rules.react.js
KatPo: yes, I was looking for `underline` rule in SimpleMarkdown instead of just `u`, so this rule is completely not needed, default one is enough

web/markdown/rules.react.js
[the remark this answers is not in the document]
Ashoat: So what happens with just a single newline? Like this:

```
> Block quote
Text
```

web/markdown/rules.react.js
Ashoat: I noticed that all of `simple-markdown` RegExes start with `^`. I'm working if matching in the middle of the text will break `simple-markdown`. I think [this line](https://github.com/Khan/simple-markdown/blob/master/simple-markdown.js#L464) assumes that the capture happens at the start of the text

lib/shared/markdown.js
```diff
@@ -5,7 +5,7 @@ const paragraphRegex = /^((?:[^\n]*)(?:\n|$))/;
 const blockQuoteRegex = /^( *>[^\n]+(?:\n[^\n]+)*)(?:\n\n|$)/;
 const headingRegex = /^ *(#{1,6}) ([^\n]+?)#* *(?:\n|$)/;
 const codeBlockRegex = /^(?: {4}[^\n]+\n*)+(?:\n *)/;
-const fenceRegex = /^(`{3,}|~{3,})(?:(.*$)\n)?([\s\S]*)\1(?:\n|$)/m;
+const fenceRegex = /^(`{3,}|~{3,})(?:(.*$)\n)?([\s\S]*)\1(?:\n|$)/;
```
KatPo, line 8: I've changed what you suggested a little bit, and added non capturing group at the end, to not include empty line that's left after \``` . I also wanted to use `m` to achieve the same with empty line at the begining, but then I've realized it's breaking in some cases

Ashoat:
We've [moved](https://phabricator.ashoat.com/D4) this PR to Phabricator! Let's follow up there

PR #107, opened 2020-07-21
Add userInfo state check

zrebcu411:


Ashoat:
We've [moved](https://phabricator.ashoat.com/D2) this PR to Phabricator! Let's follow up there

PR #210, opened 2023-01-26
path resolution plus default port

naveed949:
# Introduction
This PR includes the following two fixings:

- path resolution
- default port
# Details
1. Path resolution
I'd got a scenario where 
it was replacing a similar directory in the path instead of the target pkg: 
e.g.
 file:///Users/XXXX/`web`3Service/comm/web/media/file-utils.js  was getting resolved into file:///Users/XXXX/`keyserver/dist/web`3Service/comm/web/media/file-utils.js where `pkg = 'web'`
 
 2. added default port 3000 in `desktop/src/main.js`

### DIFF
--- a/desktop/src/main.js
+++ b/desktop/src/main.js
@@ -14,7 +14,7 @@ import path from 'path';
 import { handleSquirrelEvent } from './handle-squirrel-event';
 
 const isDev = process.env.ENV === 'dev';
-const url = isDev ? 'http://localhost/comm/' : 'https://web.comm.app';
+const url = isDev ? 'http://localhost:3000/comm/' : 'https://web.comm.app';
 const isMac = process.platform === 'darwin';
 
 const scrollbarCSS = fs.promises.readFile(
--- a/keyserver/loader.mjs
+++ b/keyserver/loader.mjs
@@ -15,7 +15,7 @@ async function resolve(specifier, context, nextResolve) {
       continue;
     }
     const path = localPackages[pkg];
-    const url = defaultResult.url.replace(path, `keyserver/dist/${pkg}`);
+    const url = defaultResult.url.replace(`comm/${path}`, `comm/keyserver/dist/${pkg}`);
     return { url };
   }
 


keyserver/loader.mjs
```diff
@@ -15,7 +15,7 @@ async function resolve(specifier, context, nextResolve) {
       continue;
     }
     const path = localPackages[pkg];
-    const url = defaultResult.url.replace(path, `keyserver/dist/${pkg}`);
+    const url = defaultResult.url.replace(`comm/${path}`, `comm/keyserver/dist/${pkg}`);
```
Ashoat, line 18: Is there a way to fix your scenario without requiring that the git repo is in a folder called `comm`?

desktop/src/main.js
```diff
@@ -14,7 +14,7 @@ import path from 'path';
 import { handleSquirrelEvent } from './handle-squirrel-event';
 
 const isDev = process.env.ENV === 'dev';
-const url = isDev ? 'http://localhost/comm/' : 'https://web.comm.app';
+const url = isDev ? 'http://localhost:3000/comm/' : 'https://web.comm.app';
```
Ashoat, line 17: I think this is a good change given the recent changes in the dev environment to no longer use Apache as a proxy.

@MichalGniadek, does this change make sense to you?

Ashoat (changes requested):
Hi @naveed949!!Thanks for submitting this :)

Requesting changes for a response on the `loader.mjs` change

desktop/src/main.js
MichalGniadek: Yeah, definitely!

keyserver/loader.mjs
naveed949: @Ashoat please check the latest commit, fixed it without `comm` :)

keyserver/loader.mjs
```diff
@@ -15,7 +15,7 @@ async function resolve(specifier, context, nextResolve) {
       continue;
     }
     const path = localPackages[pkg];
-    const url = defaultResult.url.replace(path, `keyserver/dist/${pkg}`);
+    const url = defaultResult.url.replace(specifier, `keyserver/dist/${specifier}`);
```
Ashoat, line 18: I find it really hard to believe that this works for `rust-node-addon`, and I don't understand how it solves your original issue either

Ashoat (changes requested):

keyserver/loader.mjs
naveed949: In following URL:
> file:///Users/XXXX/`web`3Service/comm/`web`/media/file-utils.js

where `path= 'web'` & `specifier= 'web/media/file-utils'`

previous logic was replacing **first** matching `path` with `keyserver/dist/${pkg}` which in my case (shared above URL) produced/resolved into the following incorrect URL:

> file:///Users/XXXX/`keyserver/dist/web`3Service/comm/web/media/file-utils.js`

THIS was my original scenario, which I've fixed by replacing `specifier` with `keyserver/dist/${specifier}` instead of `path` ( because there are very rare chances of `specifier`  duplications in a URL)

That's how given logic is resolving my issue. 
resolving this:
> file:///Users/XXXX/web3Service/comm/`web/media/file-utils`.js

into

> file:///Users/XXXX/web3Service/comm/`keyserver/dist/web/media/file-utils`.js

As far as `rust-node-addon` I can not produce it failing.

Ashoat: That explanation does not make sense. You claim that "there are very rare chances of `specifier` duplication", but if you look at the code `specifier` is exactly equal to `pkg`, and `pkg` is exactly equal to `path` for all cases other than `rust-node-addon`. So your change results in exactly the same behavior for `web`, `landing`, and `lib`. Meanwhile, it's very likely to break `rust-node-addon`.

I suspect you aren't testing your changes very well.

Ashoat:
Hi @naveed949, unfortunately my cycles are too limited here to help you through the `loader.mjs` work... you're missing some basic stuff and I worry I'll have to handhold you too much to get this PR landed.

If you can get rid of your changes to `loader.mjs` and just leave the changes to `desktop/src/main.js`, we can get this landed. Otherwise I'll probably close this PR and we can have @MichalGniadek land the `desktop/src/main.js` changes.

Ashoat:
`desktop/src/main.js` changes have been landed, and the `loader.mjs` changes still don't make sense to me – closing this

PR #329, opened 2025-12-04
[lib] support E2EE DMs only for Vodozemac clients

xsanm:
Summary:
Solution 3 from [ENG-11545](https://linear.app/comm/issue/ENG-11545/handle-olm-x3dh-bug).

Some context when to merge it in [ENG-11554](https://linear.app/comm/issue/ENG-11554/deprecate-supporting-device-with-olm).

I had to move `useGetDeviceListsForUsers` because of a dependency cycle.

Test Plan: Set `NEXT_CODE_VERSION` for something different and test what `useUsersSupportThickThreads` returns for devices during searching users

Reviewers: ashoat

Needs to be rebased on https://github.com/CommE2E/comm/pull/328

### DIFF
--- a/lib/hooks/peer-list-hooks.js
+++ b/lib/hooks/peer-list-hooks.js
@@ -3,6 +3,7 @@
 import invariant from 'invariant';
 import * as React from 'react';
 
+import { useGetDeviceListsForUsers } from './use-get-device-lists-for-users.js';
 import { setPeerDeviceListsActionType } from '../actions/aux-user-actions.js';
 import { usePeerOlmSessionsCreatorContext } from '../components/peer-olm-session-creator-provider.react.js';
 import {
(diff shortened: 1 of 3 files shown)


lib/hooks/peer-list-hooks.js
```diff
@@ -14,23 +15,21 @@ import { usePeerToPeerCommunication } from '../tunnelbroker/peer-to-peer-context
 import { useTunnelbroker } from '../tunnelbroker/tunnelbroker-context.js';
 import { useResendPeerToPeerMessages } from '../tunnelbroker/use-resend-peer-to-peer-messages.js';
 import type {
-  UsersRawDeviceLists,
-  UsersDevicesPlatformDetails,
-  SignedDeviceList,
   RawDeviceList,
+  SignedDeviceList,
```
Ashoat: Concerning to see these moved around, makes me wonder if the prior PR would fail ESLint CI

Ashoat (approved):

